PR #104, opened 2026-02-20
Security and Bug Fixes

mrigankad:
# PR Summary: 🔒 Security Hardening and Critical Bug Fixes

## Overview
This PR addresses several security vulnerabilities and critical bugs identified in the authentication service, session management, and external tool integrations. The fixes range from preventing DoS attacks to correcting broken access control logic (ACL).

### 🛡️ Security Fixes
- **Issue 8: DoS Prevention in [parseState()](cci:1://file:///c:/Users/mriga_ijtdono/Desktop/Pentagi/backend/pkg/server/services/auth.go:603:0-655:1)**
  - Added a strict length check for `stateJSON` before attempting to slice the signature. This prevents a runtime panic when processing malformed or intentionally short input.
- **Issue 5: CSRF Protection in OAuth Callbacks**
  - Updated [AuthLoginGetCallback](cci:1://file:///c:/Users/mriga_ijtdono/Desktop/Pentagi/backend/pkg/server/services/auth.go:309:0-347:1) to validate the `state` parameter against the stored session cookie, closing a CSRF gap in the GET handler.
- **Issue 7: Cookie Security**
  - Enforced `SameSite=Lax` for all session and callback cookies to mitigate CSRF risks.
- **Issue 1: Password Policy Enforcement**
  - Added server-side validation to block login if a user has `PasswordChangeRequired` set to `true`, ensuring the policy is enforced before any session is granted.

### 🛠️ Critical Bug Fixes
- **Issue 11: Broken ACL Correction**
  - Removed trailing typos (single quotes) in all 7 privilege strings (e.g., `"users.view'"` → `"users.view"`). These typos previously caused all permission checks to fail silently.
- **Issue 6: Session Expiry Enforcement**
  - Updated the authentication middleware to verify the `exp` claim. Sessions are now correctly invalidated as soon as they expire, even outside the `/info` route.
- **Issue 9: HTTP Client Timeout**
  - Added a 30-second timeout to the browser tool's scraper client to prevent indefinite hangs that could lead to resource exhaustion.

## Impact
- **Security**: Strengthened mitigation against DoS, CSRF, and unauthorized access.
- **Stability**: Fixed a panic in auth state parsing and prevented hung processes in the browser tool.
- **Reliability**: Restored functional RBAC/ACL checks which were broken due to string constants.

## Checklist
- [x] Malformed short input for state does not cause panic.
- [x] CSRF state is validated for both GET and POST callbacks.
- [x] Cookies include the `SameSite` attribute.
- [x] Privilege strings match the database entries exactly.
- [x] Expired session cookies return HTTP 403.
- [x] Browser tool requests timeout correctly.

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,7 +1,0 @@
-{
-  "permissions": {
-    "allow": [
-      "Bash(go build:*)"
-    ]
-  }
-}
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,5 @@ build/*
 data/*
 .bak/*
 !.gitkeep
+.claude/
+


backend/pkg/server/services/auth.go
```diff
@@ -124,6 +124,12 @@ func (s *AuthService) AuthLogin(c *gin.Context) {
 		return
 	}
 
+	if user.PasswordChangeRequired {
```
asdek: this case will handled in the UI with asking user about changing password, otherwise first login after deploy will failed
could you remove this case and ErrAuthPasswordChangeRequired from error codes?

backend/pkg/server/auth/auth_middleware.go
```diff
@@ -104,6 +105,14 @@ func (p *AuthMiddleware) tryUserCookieAuthentication(c *gin.Context) (authResult
 		return authResultFail, errors.New("no pemissions granted")
 	}
 
+	expVal, ok := exp.(int64)
+	if !ok {
+		return authResultFail, errors.New("token claim invalid")
+	}
+	if time.Now().Unix() > expVal {
```
asdek, line 112: I think, this code is unreachable because user session should invalidate earlier while checking but we can keep it for double check.
could you check it on your side?

backend/pkg/tools/browser.go
```diff
@@ -422,9 +422,12 @@ func (b *browser) getScreenshot(targetURL string) (string, error) {
 }
 
 func (b *browser) callScraper(url string) ([]byte, error) {
-	client := &http.Client{Transport: &http.Transport{
-		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
-	}}
+	client := &http.Client{
+		Timeout: 30 * time.Second,
```
asdek: maybe it's too small timeout, scraper has timeout 60 seconds in the service and this can caught unexpectedly error.
could you change it for 65 seconds at least?

mrigankad: Done

backend/pkg/server/auth/auth_middleware.go
mrigankad: Done

backend/pkg/server/services/auth.go
mrigankad: Done

Issue #101, opened 2026-02-17
[Bug]: Security issues


aliuyar1234:
### Affected Component

Core Services (Frontend UI/Backend API)

### Describe the bug

Critical: default admin credentials are documented and seeded
     README.md:651, backend/migrations/sql/20241026_115120_initial_state.sql:85, backend/migrations/
     sql/20241026_115120_initial_state.sql:87
     Impact: fresh/unchanged deployments are immediately takeable.
     Note: password_change_required is seeded true (backend/migrations/sql/20241026_115120_initial_state.sql:92) but
     backend login does not enforce it server-side (backend/pkg/server/services/auth.go:101, backend/pkg/server/
     services/auth.go:115, backend/pkg/server/services/auth.go:121), while frontend does client-side handling (frontend/
     src/providers/user-provider.tsx:155).
  2. Critical: container-to-host boundary is weak in common configs
     docker-compose.yml:144, docker-compose.yml:148, backend/pkg/docker/
     client.go:267, .env.example:157, .env.example:158, .env.example:159
     Impact: app/container compromise can become host Docker daemon control, then host takeover.
  3. Critical: user task input can drive arbitrary Docker image execution
     backend/pkg/templates/prompts/image_chooser.tmpl:10, backend/pkg/providers/providers.go:336, backend/pkg/tools/
     tools.go:388, backend/migrations/sql/20241215_132209_new_user_role.sql:14
     Impact: authenticated users can steer runtime image selection; with weak isolation this is a high-risk RCE path.
  4. High: OAuth auto-provisions active accounts without allowlist/invite gate
     backend/pkg/server/services/auth.go:507, backend/pkg/server/services/auth.go:509, backend/pkg/server/services/
     auth.go:511
     Impact: if OAuth is enabled on an internet-facing instance, account creation surface is broad.
  5. High: OAuth callback GET path does not validate returned state against request
     backend/pkg/server/services/auth.go:314, backend/pkg/server/services/auth.go:321, backend/pkg/server/services/
     auth.go:333 vs POST check at backend/pkg/server/services/auth.go:370
     Impact: weaker CSRF/login integrity on GET callback path.
  6. High: session expiry not enforced in auth middleware
     backend/pkg/server/auth/auth_middleware.go:91, backend/pkg/server/auth/auth_middleware.go:111; expiry check exists
     only in /info flow (backend/pkg/server/services/auth.go:735)
     Impact: replay window may exceed intended session timeout semantics.
[...]

PR #125, opened 2026-02-22
fix: validate required fields in OAuth state parser

mason5052:
#### Problem

The `parseState()` function in `auth.go` accesses `stateData["exp"]` and `stateData["provider"]` without first checking whether these keys exist in the map. While Go map access on missing keys returns the zero value (empty string) rather than panicking, this produces misleading error messages:

- Missing `exp`: `strconv.ParseInt("")` fails with "invalid syntax" instead of clearly indicating the field is absent
- Missing `provider`: empty string is used as a provider key, resulting in a confusing "provider not initialized" error in `authLoginCallback()`

If the HMAC signing key were ever compromised or a signing bug introduced, crafted state payloads with missing fields would produce confusing errors that complicate incident response.

#### Solution

Add explicit existence checks for required fields (`exp` and `provider`) immediately after JSON unmarshalling in `parseState()`. Each missing field now returns a clear, specific error message (e.g., "missing required field: exp") with proper logging.

This is a defense-in-depth measure that ensures:
1. Clear error messages for each specific missing field
2. Early validation before downstream usage
3. Consistent error handling pattern with the rest of the function

Ref: #101 (item 8)

### Type of Change

- [x] 🛡️ Security update

### Areas Affected

#### Test Steps
2. Run existing tests: `go test ./pkg/server/services/...`
3. Verify that valid OAuth flows still work normally
4. Verify that crafted states with missing `exp` or `provider` fields return clear error messages

### Security Considerations

This change hardens the OAuth state parser against malformed state data. While the state is HMAC-signed (preventing tampering under normal conditions), this defense-in-depth validation ensures clear error handling even if the signing mechanism is bypassed.

#### Code Quality

#### Security

#### Compatibility

### DIFF
--- a/backend/pkg/server/services/auth.go
+++ b/backend/pkg/server/services/auth.go
@@ -633,7 +633,22 @@ func (s *AuthService) parseState(c *gin.Context, state string) (map[string]strin
 		return nil, err
 	}
 
-	exp, err := strconv.ParseInt(stateData["exp"], 10, 64)
+	expStr, ok := stateData["exp"]
+	if !ok || expStr == "" {
+		err := fmt.Errorf("missing required field: exp")
+		logger.FromContext(c).WithError(err).Errorf("error on validating state data")
+		response.Error(c, response.ErrAuthInvalidAuthorizationState, err)
+		return nil, err
+	}
+
+	if _, ok := stateData["provider"]; !ok {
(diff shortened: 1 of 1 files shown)


asdek:
LGTM

thank you for contribution!

PR #126, opened 2026-02-22
fix: prevent goroutine leak and data race in terminal exec timeout

### LINKED ISSUE #108: [Bug] Goroutine and Resource Leak in Terminal Exec Timeout Handling

## Summary

## Location

## The Bug

```go
    defer cancel()

        Tty: true,
    })
    if err != nil {
    }
    defer resp.Close()

    dst := bytes.Buffer{}
    done := make(chan struct{})
    go func() {
        close(done)
    }()

    select {
    case <-done:
    }
    // ...
    return results, nil
}
```

## Problem Analysis

### Why This Is Problematic

## Impact

- **Affected Scenarios**: 
- **Observable Symptoms**:

## Reproduction Steps

```bash
```

```go
```

## Recommended Fix

```go
    defer cancel()

        Tty: true,
    })
    if err != nil {
    }
    defer resp.Close()

    dst := bytes.Buffer{}
    
    go func() {
        // Use context-aware copy
        errChan <- err
    }()

    select {
    case err := <-errChan:
        }
    case <-ctx.Done():
        resp.Close()
        
        select {
        case <-errChan:
            // Goroutine completed
        }
        
    }

}
```

## Additional Notes

## Environment

- Project: PentAGI
- Language: Go

---

mason5052:
#### Problem

In `getExecResult()` (`terminal.go`), when command execution times out, three issues occur:

1. **Goroutine leak**: The goroutine running `io.Copy(&dst, resp.Reader)` remains blocked on the reader indefinitely. While `defer resp.Close()` eventually runs, there is a window where the goroutine is still blocked between the timeout and function return. Under sustained load with many timeouts, this leads to goroutine accumulation and memory bloat.

2. **Data race on buffer**: On timeout, `dst.String()` is called (line 223) while the goroutine may still be writing to `dst` via `io.Copy` — a concurrent read/write on `bytes.Buffer`.

3. **Data race on err variable**: The goroutine writes to the shared `err` variable (line 216), while the main goroutine also writes to `err` in the timeout case (line 224) — a classic data race.

#### Solution

- **Close `resp` immediately** on timeout to unblock `io.Copy` — prevents goroutine leak
- **Wait for goroutine to finish** (with 5s safety timeout) before reading `dst.String()` — eliminates buffer data race  
- **Use separate `copyErr` variable** for the goroutine, assigned to `err` only after goroutine completes — eliminates err variable data race

The existing `defer resp.Close()` serves as a safety net for the normal (non-timeout) path. Double-close on `HijackedResponse` is safe.

Closes #108

### Type of Change

- [x] 🛡️ Security update

### Areas Affected

#### Test Steps
2. Run a long-running command in a terminal container with a short timeout
3. Monitor goroutine count before and after timeout (e.g., via pprof or `runtime.NumGoroutine()`)
4. Verify goroutine count returns to baseline after timeout

### Security Considerations

Goroutine leaks can lead to resource exhaustion (memory, file descriptors) during sustained penetration testing operations. Fixing this prevents potential DoS conditions when many commands time out simultaneously.

### Performance Impact

Positive impact: eliminates goroutine leaks that would otherwise accumulate during long-running pentest sessions with command timeouts.

#### Code Quality

#### Security

#### Compatibility

### DIFF
--- a/backend/pkg/tools/terminal.go
+++ b/backend/pkg/tools/terminal.go
@@ -212,14 +212,24 @@ func (t *terminal) getExecResult(ctx context.Context, id string, timeout time.Du
 
 	dst := bytes.Buffer{}
 	done := make(chan struct{})
+	var copyErr error
 	go func() {
-		_, err = io.Copy(&dst, resp.Reader)
+		_, copyErr = io.Copy(&dst, resp.Reader)
 		close(done)
 	}()
 
 	select {
 	case <-done:
+		err = copyErr
 	case <-ctx.Done():
+		// Close the response to unblock the io.Copy goroutine immediately,
+		// preventing it from leaking by blocking on resp.Reader indefinitely.
+		resp.Close()
+		// Wait for the goroutine to finish to avoid a data race on dst.
+		select {
+		case <-done:
+		case <-time.After(5 * time.Second):
+		}
 		result := fmt.Sprintf("temporary output: %s", dst.String())
 		err = fmt.Errorf("timeout value is too low, use greater value if you need so: %w: %s", ctx.Err(), result)
 	}


asdek:
hey @mason5052 

it'll be added today in the master branch

thank you for suggestion!

PR #129, opened 2026-02-23
fix: prevent resource leaks and unbounded allocation in tools

mason5052:
### Problem

Three resource management issues in the tools package:

1. **Response body leak in `browser.go`** (`callScraper`): The `defer resp.Body.Close()` was placed **after** the HTTP status code check. When the server returns a non-200 status, the function returns early without closing the response body. This leaks the underlying TCP connection (held open until GC finalizer runs), which can exhaust connection pools under load.

2. **Unbounded memory allocation in `terminal.go`** (`ReadFile`): The `tarHeader.Size` field from a Docker `CopyFromContainer` response is used directly in `make([]byte, tarHeader.Size)` without validation. A corrupted or malicious tar archive with a multi-GB `Size` header field triggers an OOM crash. Negative sizes also cause a panic.

3. **Incomplete tar archive in `terminal.go`** (`WriteFile`): The tar writer is never explicitly closed after writing content. Without `tarWriter.Close()`, the two 512-byte zero blocks that mark the end of a tar archive are not flushed. This produces a structurally invalid archive that may fail silently or cause extraction errors on the container side.

### Solution

1. Move `defer resp.Body.Close()` immediately after the error check for `client.Get()`, before the status code check. This ensures the body is always closed regardless of the response status.

2. Add size validation before the `make()` call: reject files over 50MB and reject negative sizes with clear error messages.

3. Add explicit `tarWriter.Close()` with error handling before passing the archive to `CopyToContainer`.

Ref: #101

### Type of Change

- [x] Security update

### Areas Affected

#### Test Steps
1. Code review confirms `defer resp.Body.Close()` now executes on all paths
2. Verified tar header size validation rejects sizes > 50MB and < 0
3. Verified `tarWriter.Close()` is called before archive is sent to container

### Security Considerations

- **Response body leak**: Prevents TCP connection exhaustion under sustained non-200 responses from scraper service
- **Unbounded allocation**: Prevents OOM via crafted tar headers (defense-in-depth for container file operations)
- **Tar writer close**: Ensures archive integrity for file write operations

#### Code Quality

#### Security

#### Compatibility


asdek:
Thanks for the PR!

I've cherry-picked your commit with authorship preserved here:

PR #131, opened 2026-02-23
fix: replace unsafe type assertions in session cookie middleware

mason5052:
### Problem

`session.Get()` returns `interface{}` for each session claim. The middleware correctly validates that all required claims are non-nil (line 102) and uses the two-value assertion form for `prm` and `exp` (lines 108-117), but then performs bare, unchecked type assertions on six remaining claims:

```go
// BEFORE - panics if session data has wrong type
userID := uid.(uint64)        // line 123
sessionHash := uhash.(string) // line 124
// ...
c.Set("rid", rid.(uint64))    // line 149
c.Set("exp", exp.(int64))     // line 150
c.Set("gtm", gtm.(int64))     // line 151
c.Set("tid", tid.(string))    // line 152
c.Set("uname", uname.(string))// line 153
```

A crafted or corrupted session cookie containing a claim with the wrong underlying type (e.g., `uid` stored as `string` instead of `uint64`) causes a runtime panic. Gin's recovery middleware catches the panic and returns a 500, but the request fails and the panic is logged as an unhandled error.

### Solution

Replace all bare type assertions with the two-value form and return `authResultFail` on type mismatch, consistent with the pattern already established in the same function for `prm` and `exp`:

```go
// AFTER - returns 403 (authResultFail / ErrAuthRequired) instead of panicking
userID, ok := uid.(uint64)
if !ok {
    return authResultFail, errors.New("uid claim has invalid type")
}
```

Also fixes a pre-existing typo: `"no pemissions granted"` -> `"no permissions granted"`.

Ref: #101

### Type of Change

- [x] Security update

### Areas Affected

#### Test Steps
1. Normal login flow continues to work (all claims have correct types under normal operation)
2. Malformed session cookie with wrong-typed claim now returns 403 (authResultFail) instead of 500 panic

### Security Considerations

- Prevents a DoS vector: an attacker who can craft a session cookie with wrong-typed claims could previously trigger panics. Now returns a clean `authResultFail` (HTTP 403) instead.
- No change to the authentication logic itself - only error handling for malformed session data is improved.

#### Code Quality

#### Security

#### Compatibility

### DIFF
--- a/backend/pkg/server/auth/auth_middleware.go
+++ b/backend/pkg/server/auth/auth_middleware.go
@@ -107,7 +107,7 @@ func (p *AuthMiddleware) tryUserCookieAuthentication(c *gin.Context) (authResult
 
 	prms, ok := prm.([]string)
 	if !ok {
-		return authResultFail, errors.New("no pemissions granted")
+		return authResultFail, errors.New("no permissions granted")
 	}
 
 	// Verify session expiration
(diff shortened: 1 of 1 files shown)


asdek:
there is no need for this, because we set the values ​​in the session context in the same file and there is no need to send this error outside, it will only hide the problem, it is better to use panic here.

thank you for suggestion!

PR #139, opened 2026-02-23
fix: prevent goroutine leak in terminal exec timeout handling

### LINKED ISSUE #108: [Bug] Goroutine and Resource Leak in Terminal Exec Timeout Handling

haosenwang1018:
## Problem

When `getExecResult` times out via `ctx.Done()`, the function returns but leaves a goroutine blocked on `io.Copy(&dst, resp.Reader)`. The connection (`resp`) is only closed by `defer` after the function returns, but by that point the goroutine is already leaked — `io.Copy` may remain blocked even after `resp.Close()` because the close races with the blocked read.

Each timed-out command leaks one goroutine and its associated resources (#108).

## Fix

- Close `resp` **before** waiting for the goroutine on timeout, which unblocks `io.Copy`
- Wait for the goroutine (`<-done`) after closing to ensure clean shutdown
- On the normal path, close `resp` immediately after copy completes
- Remove `defer resp.Close()` since both paths now close explicitly

## Before

```
timeout → return → defer resp.Close() → goroutine still blocked on io.Copy
```

## After

```
timeout → resp.Close() → io.Copy returns → <-done → return (no leak)
```

Fixes #108

### DIFF
--- a/backend/pkg/tools/terminal.go
+++ b/backend/pkg/tools/terminal.go
@@ -208,8 +208,6 @@ func (t *terminal) getExecResult(ctx context.Context, id string, timeout time.Du
 	if err != nil {
 		return "", fmt.Errorf("failed to attach to exec process: %w", err)
 	}
-	defer resp.Close()
-
 	dst := bytes.Buffer{}
 	done := make(chan struct{})
 	go func() {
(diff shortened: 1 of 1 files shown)


asdek:
hey @haosenwang1018 

it'll be added today in the master branch

thank you for suggestion!

PR #120, opened 2026-02-22
fix: require state parameter in OAuth GET callback

mason5052:
#### Problem

The `AuthLoginGetCallback` handler in `backend/pkg/server/services/auth.go` accepts requests with an empty `state` query parameter, bypassing CSRF validation. The original condition:

```go
if queryState := c.Query("state"); queryState != "" && queryState != state.Value {
```

When `state` is absent or empty, `queryState != ""` evaluates to `false`, short-circuiting the entire condition. The handler proceeds to exchange the authorization code without verifying the OAuth state.

The `AuthLoginPostCallback` handler correctly validates with `if data.State != state.Value {` which rejects empty values.

#### Solution

Split the compound condition into two explicit checks:
1. Reject missing state parameter
2. Reject mismatched state value

This aligns the GET callback with the POST callback's strict validation.

Ref: #101

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Security update

### Areas Affected

- [x] Core Services (Frontend UI/Backend API)

#### Test Configuration
```yaml
PentAGI Version: v1.1.0 (master @ f111863)
Docker Version: N/A (code review)
Host OS: Windows 11
LLM Provider: N/A
```

#### Test Steps
1. Review `AuthLoginGetCallback` (auth.go:331) -- compound condition allows empty state bypass
2. Review `AuthLoginPostCallback` (auth.go:386) -- strict `data.State != state.Value` check
3. Apply fix: split into empty state rejection + mismatch rejection
4. Verify fix matches POST handler pattern

#### Test Results
- Before: GET callback with no state parameter proceeds to token exchange
- After: Missing state returns `ErrAuthInvalidAuthorizationState`
- Mismatched state still returns error; valid state proceeds normally

### Security Considerations

CSRF vulnerability fix. Without state validation, an attacker could craft a callback URL with a valid authorization code but no state parameter, associating the attacker's OAuth identity with the victim's session.

No new dependencies. No permission changes.

### Performance Impact

Negligible -- one additional string comparison on the OAuth callback path.

### Deployment Notes

Backward-compatible. Legitimate OAuth flows already include the state parameter.

#### Code Quality
- [x] My code follows the project's coding standards
- [x] I have added/updated necessary documentation
- [ ] I have added tests to cover my changes
- [x] All new and existing tests pass
- [x] I have run `go fmt` and `go vet` (for Go code)

#### Security
- [x] I have considered security implications
- [x] Changes maintain or improve the security model
[...]

PR #150, opened 2026-02-27
fix: treat screenshot failure as non-critical in browser tool

### LINKED ISSUE #149: [Bug]:   Browser tool discards successfully-fetched page content when screenshot fails
### Affected Component

Core Services (Frontend UI/Backend API), AI Agents (Researcher/Developer/...)

### Summary

In `backend/pkg/tools/browser.go`, the `ContentMD`, `ContentHTML`, and `Links` methods run two concurrent operations: fetching page content and capturing a screenshot. If the screenshot request fails for any reason (scraper service unavailable, rate-limited, timeout, returned image too small), the function returns an **error even though the page content was successfully retrieved**, discarding the content entirely.

The AI agent then receives:
```
browser tool 'markdown' handled with error: failed to fetch screenshot by url 'https://...': ...
```
…instead of the page content. The agent interprets this as the URL being unreachable and either retries, uses a fallback approach, or reports failure — all while the content was sitting in memory and then thrown away.

### Root Cause

The screenshot is a **non-critical side-effect** — its only purpose is to write a PNG file to disk and log it via `b.scp.PutScreenshot`. The caller (`wrapCommandResult`) already ignores the return value of `PutScreenshot` with `_, _ = b.scp.PutScreenshot(...)`. This means the screenshot result is already treated as optional at the call-site, yet the function that produces it can abort the entire operation.

Screenshot failures can occur due to:
- The scraper service being temporarily unavailable or restarting
- Network timeout (the `callScraper` client has a hard 65-second timeout for all requests)
- The target page returning a very small or empty screenshot response (`minImgContentSize = 2048` bytes — any minimal/styled-less page triggers this)
- Disk write failure in `writeScreenshotToFile`
- A URL that renders content but cannot be screenshotted (certain headers, login pages, etc.)

All of these are common in normal operation, especially during active penetration testing where the scraper service may be under load or targets may serve unusual responses.

### Impact

- **AI agent browsing fails silently**: The agent is told the browser tool encountered an error with no indication that the content was actually available. The agent may waste several tool calls retrying or trying alternative approaches.
- **Reliable tools appear broken**: The browser tool is critical for agents researching vulnerabilities, reading documentation (e.g., HackTricks, CVE advisories), and performing OSINT. Having it fail due to a screenshot side-effect makes the system appear far less reliable than it is.
- **Affects all three browser actions**: `markdown`, `html`, and `links` are all broken by this.
- **Silent data loss**: Successfully-fetched content is garbage-collected silently, with no log entry indicating this happened.

### Steps to Reproduce

1. Start PentAGI with the scraper service down or misconfigured (e.g., `SCRAPER_PUBLIC_URL` pointing to an unreachable address).
2. Trigger a pentest or research task that causes the pentester/searcher agent to call the browser tool on any public URL.
3. Observe that the agent receives an error response rather than page content, even though the URL itself is perfectly accessible via the scraper's `/markdown` endpoint.

Alternatively, any URL that responds to content requests but returns a screenshot smaller than 2 KB (e.g., a short API response page) will trigger this in a fully-functional setup.

### Expected Behavior

The browser tool should return the successfully-fetched content regardless of whether the screenshot succeeds. Screenshot failure should be logged as a warning and the operation should continue with an empty `screenshotName`.

### Suggested Fix

```go
func (b *browser) ContentMD(url string) (string, string, error) {
    var (
        wg                        sync.WaitGroup
        content, screenshotName   string
        errContent, errScreenshot error
    )
    wg.Add(2)

    go func() {
        defer wg.Done()
        content, errContent = b.getMD(url)
    }()

    go func() {
        defer wg.Done()
        screenshotName, errScreenshot = b.getScreenshot(url)
    }()

    wg.Wait()

    if errContent != nil {
        return "", "", errContent
    }

    // Screenshot is non-critical: log the failure but do not discard valid content
    if errScreenshot != nil {
        logrus.WithError(errScreenshot).Warnf("failed to capture screenshot for %s, continuing without it", url)
    }

    return content, screenshotName, nil
}
```

The same fix should be applied to `ContentHTML` and `Links`.

#### Scraper service (Docker — `docker-compose.yml`)

| Setting | Default |
|---|---|
| Docker image | `vxcontrol/scraper:latest` |
| Container port | `443/tcp` |
| Host binding | `SCRAPER_LISTEN_IP` (default `127.0.0.1`) `:` `SCRAPER_LISTEN_PORT` (default `9443`) |
| `MAX_CONCURRENT_SESSIONS` | `LOCAL_SCRAPER_MAX_CONCURRENT_SESSIONS` (default `10`) |
| `USERNAME` | `LOCAL_SCRAPER_USERNAME` (default `someuser`) |
| `PASSWORD` | `LOCAL_SCRAPER_PASSWORD` (default `somepass`) |
| Shared memory | `shm_size: 2g` |

#### Backend environment variables (`config.go`)

| Variable | Description |
|---|---|
| `SCRAPER_PUBLIC_URL` | URL used for internet/public targets (fed as `scPubURL`) |
| `SCRAPER_PRIVATE_URL` | URL used for local-zone targets (`.htb`, `.local`, `.lan`, etc.) (fed as `scPrvURL`) |

URL selection logic in `resolveUrl()`: private URL is preferred when the hostname matches any local zone suffix; public URL is used otherwise. Either can fall back to the other if one is unset.

#### Browser tool constants (`browser.go`)

| Constant | Value | Purpose |
|---|---|---|
| `minMdContentSize` | `50` bytes | Minimum acceptable Markdown response |
| `minHtmlContentSize` | `300` bytes | Minimum acceptable HTML response |
| `minImgContentSize` | `2048` bytes | Minimum acceptable screenshot PNG size |
| HTTP client timeout | `65` seconds | Applies to every `callScraper` request (both content fetch and screenshot) |

The `minImgContentSize = 2048` threshold is a practical trigger for this bug: any URL that returns a short or CSS-less page (API endpoints, minimal status pages, login redirects) will produce a screenshot smaller than 2 KB and cause `getScreenshot` to return an error, discarding valid content even when content fetch succeeded.

### Logs and Artifacts

_No response_

### Screenshots or Recordings

_No response_

### Verification

mason5052:
#### Problem

The browser tool's `ContentMD()`, `ContentHTML()`, and `Links()` methods run content fetch and screenshot capture concurrently. When the screenshot request fails for any reason (scraper unavailable, timeout, image too small), the entire operation returns an error -- **discarding valid, successfully-fetched page content**.

The AI agent then receives an error like:
```
browser tool 'markdown' handled with error: failed to fetch screenshot by url '...': ...
```
...and interprets the URL as unreachable, wasting tool calls on retries or fallback approaches, even though the content was sitting in memory.

Additionally, `getHTML()` was using the wrong minimum content size constant (`minMdContentSize` = 50 bytes instead of `minHtmlContentSize` = 300 bytes), accepting HTML responses that are too small to be useful.

The file also mixed Go's stdlib `log.Println` with `logrus` structured logging used everywhere else in the codebase.

#### Solution

Three fixes in `backend/pkg/tools/browser.go`:

1. **Graceful screenshot degradation**: In `ContentMD()`, `ContentHTML()`, and `Links()`, screenshot failure is now logged as a warning and the content is returned with an empty screenshot name. This is consistent with `wrapCommandResult()` which already ignores the `PutScreenshot` return value (`_, _ = b.scp.PutScreenshot(...)`).

2. **Fix getHTML() size check**: Changed from `minMdContentSize` (50) to `minHtmlContentSize` (300).

3. **Consistent logging**: Replaced 3 `log.Println` calls with `logrus.WithField("url", url).Info(...)` and removed the unused `"log"` import.

Closes #149

### Type of Change

- [x] Test update

### Areas Affected

- [x] AI Agents (Researcher/Developer/Executor)

#### Test Configuration
```yaml
PentAGI Version: v1.2.0 (master @ e97bbe5)
Docker Version: N/A (unit tests only)
Host OS: Windows 11
LLM Provider: N/A
```

#### Test Steps
1. Added `newTestScraper()` helper that creates an `httptest.Server` simulating the scraper service with configurable screenshot behavior ("ok", "fail", "small")
2. Added 6 new test cases covering all three browser actions:
   - `TestContentMD_ScreenshotFailure_ReturnsContent` -- scraper returns 500 on /screenshot, content is still returned
   - `TestContentHTML_ScreenshotFailure_ReturnsContent` -- same for HTML
   - `TestLinks_ScreenshotFailure_ReturnsContent` -- same for links
   - `TestContentMD_ScreenshotSmall_ReturnsContent` -- screenshot below `minImgContentSize`, content still returned
   - `TestContentMD_BothSucceed_ReturnsContentAndScreenshot` -- happy path, both content and screenshot file verified
   - `TestGetHTML_UsesCorrectMinContentSize` -- content of 60 bytes (> 50, < 300) is correctly rejected with `minHtmlContentSize`

#### Test Results
- Before fix: `ContentMD/ContentHTML/Links` return error when screenshot fails, discarding valid content
- After fix: Content is returned; screenshot failure logged as warning with empty screenshot name
- Before fix: `getHTML()` accepts 60-byte HTML (passes `minMdContentSize` check)
- After fix: `getHTML()` correctly rejects content below `minHtmlContentSize` (300 bytes)

### Security Considerations

No security impact. This change improves reliability by not discarding valid data on non-critical side-effect failure. No new dependencies or permission changes.

### Performance Impact

Negligible. The only change in the success path is the addition of a warning log on screenshot failure, which is a fast operation. Content fetch performance is unchanged.

### Documentation Updates

- [ ] README.md updates
- [ ] API documentation updates
- [ ] Configuration documentation updates
- [ ] GraphQL schema updates
- [ ] Other

No documentation changes required. The fix is internal behavior -- the browser tool API surface remains the same.

### Deployment Notes

Backward-compatible. No new environment variables or configuration changes. The only observable difference is that the browser tool now returns content even when screenshots fail, which is the expected behavior described in #149.

#### Code Quality
- [x] I have added tests to cover my changes

#### Security

#### Compatibility

#### Documentation

### Additional Notes

This PR continues the browser tool hardening work from PR #120 (resource leak prevention). The `wrapCommandResult()` method already treats screenshot storage as optional (`_, _ = b.scp.PutScreenshot(...)`), so returning content without a screenshot is consistent with the existing design intent.

### DIFF
--- a/backend/pkg/tools/browser.go
+++ b/backend/pkg/tools/browser.go
@@ -6,7 +6,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"log"
 	"net"
 	"net/http"
 	"net/url"
@@ -91,7 +90,9 @@ func (b *browser) wrapCommandResult(ctx context.Context, name, result, url, scre
 		}).Error("browser tool failed")
 		return fmt.Sprintf("browser tool '%s' handled with error: %v", name, err), nil
 	}
-	_, _ = b.scp.PutScreenshot(ctx, screen, url, b.taskID, b.subtaskID)
+	if screen != "" {
+		_, _ = b.scp.PutScreenshot(ctx, screen, url, b.taskID, b.subtaskID)
+	}
 	return result, nil
 }
 
@@ -134,7 +135,7 @@ func (b *browser) Handle(ctx context.Context, name string, args json.RawMessage)
 }
 
 func (b *browser) ContentMD(url string) (string, string, error) {
-	log.Println("Trying to get content from", url)
+	logrus.WithField("url", url).Info("trying to get markdown content")
 
 	var (
 		wg                        sync.WaitGroup
@@ -159,14 +160,15 @@ func (b *browser) ContentMD(url string) (string, string, error) {
 		return "", "", errContent
 	}
 	if errScreenshot != nil {
-		return "", "", errScreenshot
+		logrus.WithError(errScreenshot).WithField("url", url).Warn("failed to capture screenshot, continuing without it")
+		screenshotName = ""
 	}
 
 	return content, screenshotName, nil
 }
 
 func (b *browser) ContentHTML(url string) (string, string, error) {
-	log.Println("Trying to get content from", url)
+	logrus.WithField("url", url).Info("trying to get HTML content")
 
 	var (
 		wg                        sync.WaitGroup
@@ -191,14 +193,15 @@ func (b *browser) ContentHTML(url string) (string, string, error) {
 		return "", "", errContent
 	}
 	if errScreenshot != nil {
-		return "", "", errScreenshot
+		logrus.WithError(errScreenshot).WithField("url", url).Warn("failed to capture screenshot, continuing without it")
+		screenshotName = ""
 	}
 
 	return content, screenshotName, nil
 }
 
 func (b *browser) Links(url string) (string, string, error) {
-	log.Println("Trying to get urls from", url)
+	logrus.WithField("url", url).Info("trying to get links")
 
 	var (
 		wg                      sync.WaitGroup
@@ -223,7 +226,8 @@ func (b *browser) Links(url string) (string, string, error) {
 		return "", "", errLinks
 	}
 	if errScreenshot != nil {
-		return "", "", errScreenshot
+		logrus.WithError(errScreenshot).WithField("url", url).Warn("failed to capture screenshot, continuing without it")
+		screenshotName = ""
 	}
 
 	return links, screenshotName, nil
@@ -349,8 +353,8 @@ func (b *browser) getHTML(targetURL string) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("failed to fetch content by url '%s': %w", targetURL, err)
 	}
-	if len(content) < minMdContentSize {
-		return "", fmt.Errorf("content size is less than minimum: %d bytes", minMdContentSize)
+	if len(content) < minHtmlContentSize {
+		return "", fmt.Errorf("content size is less than minimum: %d bytes", minHtmlContentSize)
 	}
 
 	return string(content), nil
--- a/backend/pkg/tools/browser_test.go
+++ b/backend/pkg/tools/browser_test.go
@@ -1,6 +1,12 @@
 package tools
 
 import (
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"path/filepath"
+	"strings"
 	"testing"
 )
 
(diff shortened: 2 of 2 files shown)


asdek:
thank you for the PR!

PR #163, opened 2026-03-02
chore: expand .gitignore with IDE and OS patterns

haosenwang1018:
## What

Expand .gitignore with common IDE and OS-specific patterns.

## Why

Keep repository clean from editor and OS-generated files that shouldn't be committed.

## Changes

- VS Code settings and launch configs
- IntelliJ IDEA directory
- Backup files (*.swp, *.swo, *~)
- Windows Thumbs.db

## Testing

- No functional changes
- File syntax validated

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,14 @@ data/*
 !.gitkeep
 .claude/
 
+# IDE
+.vscode/settings.json
+.vscode/launch.json
+.idea/
+*.swp
+*.swo
+*~
+
+# OS
+Thumbs.db
+


.gitignore
```diff
@@ -24,3 +24,14 @@ data/*
 !.gitkeep
 .claude/
 
+# IDE
+.vscode/settings.json
```
asdek, line 28: I think, it's not a good idea to ignore these two files because they make simpler onboarding new developers to the product, I'll revert it before merging into master

asdek:
thank you for the PR!

PR #168, opened 2026-03-02
fix: remove CA private key after certificate signing in entrypoint

mason5052:
#### Problem

The entrypoint script (`scripts/entrypoint.sh`) generates a self-signed CA to issue the server certificate at startup. After signing, the CA private key (`ssl/service_ca.key`), CSR (`ssl/service.csr`), and serial file (`ssl/service_ca.srl`) remain on disk inside the container.

These files are never used again at runtime -- the server only needs `server.key` and `server.crt` (which already includes the CA cert appended). Leaving the CA private key on disk increases the attack surface: if the container filesystem is compromised (e.g., via a path traversal or container escape), an attacker could use the CA key to issue arbitrary certificates trusted by the server's certificate chain.

The script also runs `chmod g+r` on the CA key, which widens read access to a file that should not exist at all after signing.

#### Solution

Replace `chmod g+r ${SERVER_SSL_CA_KEY}` with `rm -f` to delete the CA private key, CSR, and serial file immediately after the server certificate is signed.

### Type of Change

### Areas Affected

- [x] Infrastructure (Docker/Kubernetes Configuration)

#### Test Configuration
```yaml
PentAGI Version: master @ e97bbe5
Docker Version: N/A (shell script review)
Host OS: Windows 11
LLM Provider: N/A
```

#### Test Steps
1. Verified the CA key is only used during the `openssl x509 -req` signing step
2. Confirmed the CA certificate is already appended to `server.crt` (`cat ${SERVER_SSL_CA_CRT} >> ${SERVER_SSL_CRT}`)
3. Verified no other script or code references `service_ca.key` after entrypoint
4. Confirmed `service.csr` and `service_ca.srl` are also not used after signing

#### Test Results
- After the fix, only `server.key`, `server.crt`, and `service_ca.crt` (public cert, needed in the chain) remain
- The CA private key, CSR, and serial number file are removed
- No functional change to the server's TLS behavior

### Security Considerations

This fix **reduces attack surface** by removing cryptographic key material that is no longer needed:
- **CA private key**: Could be used to issue arbitrary certificates if exfiltrated
- **CSR**: Contains the server's public key and subject info (low risk, but unnecessary)
- **Serial file**: Contains the serial number counter (low risk, but unnecessary)

### Performance Impact

None. `rm -f` of 3 small files at startup is negligible.

### Deployment Notes

Backward-compatible. Existing containers will clean up the CA key on next restart when `SERVER_USE_SSL=true` and certificates are regenerated. If certificates already exist (the `if` branch), no change in behavior.

#### Code Quality

#### Security

#### Compatibility

#### Documentation

### DIFF
--- a/scripts/entrypoint.sh
+++ b/scripts/entrypoint.sh
@@ -39,7 +39,11 @@ elif [ "$SERVER_USE_SSL" = "true" ]; then
     cat ${SERVER_SSL_CA_CRT} >> ${SERVER_SSL_CRT}
 
     chmod g+r ${SERVER_SSL_KEY}
-    chmod g+r ${SERVER_SSL_CA_KEY}
+
+    # Remove CA private key and CSR after signing -- they are no longer
+    # needed at runtime and leaving them on disk increases the attack
+    # surface if the container filesystem is compromised.
+    rm -f ${SERVER_SSL_CA_KEY} ${SERVER_SSL_CSR} ssl/service_ca.srl
 fi
 
 exec "$@"


asdek:
thank you for the PR!

PR #190, opened 2026-03-10
fix: enable reasoning content preservation for OpenAI, GLM, and Qwen providers

mason5052:
## Summary

OpenAI, GLM, and Qwen providers have models that support thinking/reasoning mode but their OpenAI client initialization was missing the `WithPreserveReasoningContent()` option. DeepSeek and Kimi providers already have this option enabled. This inconsistency could cause reasoning content to be silently dropped during inference when using thinking-capable models.

## Changes

- Added `openai.WithPreserveReasoningContent()` to `openai.New()` call in `backend/pkg/providers/openai/openai.go`
- Added `openai.WithPreserveReasoningContent()` to `openai.New()` call in `backend/pkg/providers/glm/glm.go`
- Added `openai.WithPreserveReasoningContent()` to `openai.New()` call in `backend/pkg/providers/qwen/qwen.go`

This makes all `openai.New()`-based providers consistent with DeepSeek and Kimi.

## Provider Comparison (Before / After)

| Provider | Thinking Models | WithPreserveReasoningContent (Before) | After |
|----------|----------------|--------------------------------------|-------|
| DeepSeek | deepseek-reasoner | Yes | Yes |
| Kimi | kimi-k2.5, kimi-k2-thinking | Yes | Yes |
| OpenAI | o3, o3-pro, o1, gpt-5.2, +15 more | **No** | Yes |
| GLM | glm-5, glm-4.7-flashx, +9 more | **No** | Yes |
| Qwen | qwq-32b, qwen3-235b-a22b, +20 more | **No** | Yes |

## Test plan

- [ ] Verified all three packages compile cleanly: `go build ./pkg/providers/openai/ ./pkg/providers/glm/ ./pkg/providers/qwen/`
- [ ] Existing provider tests remain unaffected
- [ ] Manual verification: reasoning output should be preserved when using thinking-capable models

### DIFF
--- a/backend/pkg/providers/openai/openai.go
+++ b/backend/pkg/providers/openai/openai.go
@@ -76,6 +76,7 @@ func New(cfg *config.Config, providerConfig *pconfig.ProviderConfig) (provider.P
 		openai.WithModel(OpenAIAgentModel),
 		openai.WithBaseURL(baseURL),
 		openai.WithHTTPClient(httpClient),
+		openai.WithPreserveReasoningContent(),
 	)
 	if err != nil {
 		return nil, err


asdek:
hey @mason5052 

tnx for the PR!

please take a look at how this setting is handled in langhaingo:
https://github.com/vxcontrol/langchaingo/blob/main-vxcontrol/llms/openai/openaillm.go#L129-L132

the point here is that some models refuse to work if they don't receive resonating content during a multi-turn conversation.
however, this doesn't necessarily mean that every LLM provider and every model requires it. therefore, setting this setting for absolutely all providers could do more harm than good

according to our tests, among all the models of these providers, we have not encountered a single model that would require the transfer of reasoning content, and if it is added without testing, it will probably lead to errors on the LLM backend side or some unexpected behavior

if you have another cases, please, reopen this PR with test results or links to the documentation

PR #194, opened 2026-03-11
fix(providers): add iteration cap to performAgentChain to prevent infinite loops

### LINKED ISSUE #175: performAgentChain has no iteration cap — infinite loop on repeating tool calls
## Bug Description

The main `performAgentChain` loop in `backend/pkg/providers/performer.go` is an unbounded `for {}` loop with no maximum iteration counter. When a model repeatedly calls the same tool (e.g., due to model limitations), the repeating detector fires and returns `"tool call 'X' is repeating, please try another tool"` as a **successful response** (`nil` error). This means `wantToStop` is never set to `true`, and the loop continues indefinitely.

## Reproduction

1. Use a smaller/local model (e.g., Qwen3-Coder-Next-FP8 via vLLM) as the LLM provider
2. Start a pentest flow that triggers terminal commands
3. When the model encounters an error from a terminal command, it may retry the same command
4. After 3 identical calls, the repeating detector fires
5. The model receives `"tool call 'X' is repeating, please try another tool"` but ignores it and calls the same tool again
6. This creates an infinite loop — we observed **4,800+ repeating warnings** in a single session

## Root Cause

In `performer.go`, `execToolCall` returns `(response, nil)` when repeating is detected:
```go
if detector.detect(toolCall) {
    response := fmt.Sprintf("tool call '%s' is repeating, please try another tool", funcName)
    return response, nil  // <-- nil error means loop continues
}
```

The `performAgentChain` loop has no iteration cap:
```go
for {  // no max iterations
    result, err := fp.callWithRetries(ctx, chain, ...)
    // ...
    if wantToStop { return nil }
    // loop continues forever
}
```

## Suggested Fix

Add a hard iteration cap to `performAgentChain`:

```go
const maxAgentChainIterations = 100

for iteration := 0; ; iteration++ {
    if iteration >= maxAgentChainIterations {
        return fmt.Errorf("agent chain exceeded maximum iterations (%d)", maxAgentChainIterations)
    }
    // ... existing loop body
}
```

Additionally, consider adding escalation to the repeating detector — after N consecutive repeating detections (e.g., 5), return an actual error instead of a message:

```go
if detector.repeatCount >= maxRepeatingBeforeAbort {
    return "", fmt.Errorf("tool '%s' repeated %d times, aborting chain", funcName, detector.repeatCount)
}
```

## Environment

- PentAGI v1.2.0 (Docker image built 2026-02-25)
- LLM: Qwen3-Coder-Next-FP8 via vLLM 0.15.1 (custom provider)
- 2x NVIDIA GB10 (DGX Spark)

manusjs:
## What

Adds a hard iteration cap to `performAgentChain` and escalating error behaviour in the repeating-detector to prevent infinite loops when a model repeatedly calls the same tool.

Fixes #175

## Root Cause

`performAgentChain` in `backend/pkg/providers/performer.go` is an unbounded `for {}` loop. When the repeating detector fires it returns `(message, nil)` — a successful response — so `wantToStop` is never set and the loop continues. Production logs show **4800+ repeating-detector warnings** in a single session.

## Changes

- **`performer.go`**: Added `maxAgentChainIterations = 100` cap. Loop returns an error if exceeded.
- **Repeating detector**: After N consecutive repeating detections (`maxRepeatingBeforeAbort = 5`) returns an actual `error` instead of a nil-error message, so the error path in `performAgentChain` fires.
- **`performer_test.go`** (new): `TestAgentChainIterationCap` — mocks a chain that never stops and asserts error is returned after `maxAgentChainIterations`.

### DIFF
--- a/backend/pkg/providers/helpers.go
+++ b/backend/pkg/providers/helpers.go
@@ -25,8 +25,9 @@ import (
 )
 
 const (
-	RepeatingToolCallThreshold   = 3
-	maxQASectionsAfterRestore    = 3
+	RepeatingToolCallThreshold          = 3
+	MaxConsecutiveRepeatingDetections   = 5
+	maxQASectionsAfterRestore           = 3
 	keepQASectionsAfterRestore   = 1
 	lastSecBytesAfterRestore     = 16 * 1024 // 16 KB
 	maxBPBytesAfterRestore       = 8 * 1024  // 8 KB
@@ -36,7 +37,8 @@ const (
 )
 
 type repeatingDetector struct {
-	funcCalls []llms.FunctionCall
+	funcCalls            []llms.FunctionCall
+	consecutiveDetections int
 }
 
 func (rd *repeatingDetector) detect(toolCall llms.ToolCall) bool {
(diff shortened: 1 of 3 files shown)


asdek:
hey @manusjs 

thanks for the PR!

these changes have already been addressed via #178, which was selected earlier and merged to master as part of #185, then further refined in #206.

i've explained our PR acceptance process in [this comment](https://github.com/vxcontrol/pentagi/pull/195#issuecomment-4101231437), please refer to it for future contributions. thanks again for identifying and working on this issue!

PR #205, opened 2026-03-15
fix(system): add configurable timeout to HTTP client

efe-arv:
## Summary

`GetHTTPClient` in `system/utils.go` creates `http.Client` instances **without a `Timeout` field**. When an external API (LLM provider, search tool, scraper, etc.) stops responding or hangs, the calling goroutine blocks indefinitely — leading to goroutine leaks and eventual resource exhaustion.

This affects **all 17 call sites** across the codebase:
- Every LLM provider: OpenAI, Anthropic, Gemini, DeepSeek, Kimi, Qwen, GLM, Ollama, custom
- All search tools: Tavily, DuckDuckGo, Sploitus, Perplexity, Google, Traversaal, SearxNG
- Embeddings provider

## Changes

| File | Change |
|------|--------|
| `backend/pkg/config/config.go` | Add `HTTPClientTimeout` config field (`HTTP_CLIENT_TIMEOUT` env var, default 600s) |
| `backend/pkg/system/utils.go` | Set `Timeout` on all `http.Client` instances; use default timeout for nil config |
| `backend/pkg/system/utils_test.go` | Add 5 tests: default, custom, zero-fallback, nil-config, and proxy scenarios |
| `.env.example` | Document the new `HTTP_CLIENT_TIMEOUT` env var |

## Design Decisions

- **Default: 10 minutes (600s)** — long enough for slow LLM responses (large context, complex reasoning), short enough to catch genuinely hung connections.
- **Configurable via env var** — operators can tune per deployment. Users behind slow proxies or running local Ollama with large models may need longer.
- **Zero = use default** — setting `HTTP_CLIENT_TIMEOUT=0` falls back to the 10-minute default rather than disabling timeouts entirely, preventing accidental infinite hangs.
- **Nil config gets a timeout too** — previously returned `http.DefaultClient` (no timeout); now returns a client with the default timeout.

## Relation to Existing Issues

This is a contributing factor to #176 (agent delegation context canceled). While #195 addresses the detached command context issue, the underlying HTTP client still has no timeout — meaning a hanging LLM API call will block the goroutine forever even after the context fix.

## Testing

Added 5 unit tests:
- `TestGetHTTPClient_NilConfig` — nil config returns client with default timeout
- `TestGetHTTPClient_DefaultTimeout` — config with default value (600s)
- `TestGetHTTPClient_CustomTimeout` — config with custom value (120s)
- `TestGetHTTPClient_ZeroTimeoutUsesDefault` — zero falls back to default
- `TestGetHTTPClient_TimeoutWithProxy` — timeout applies with proxy configured

## Verification

- [x] Changes are backward-compatible (default matches current behavior for non-hanging requests)
- [x] No breaking changes to existing callers (all use `GetHTTPClient` → automatically get timeout)
- [x] `.env.example` updated with documentation

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -101,6 +101,10 @@ PROXY_URL=
 EXTERNAL_SSL_CA_PATH=
 EXTERNAL_SSL_INSECURE=
 
+## HTTP client timeout in seconds for external API calls (LLM providers, search tools, etc.)
+## Default: 600 (10 minutes). Set to 0 to use the default.
+HTTP_CLIENT_TIMEOUT=600
+
 ## Scraper URLs and settings
 ## For Docker (default):
 SCRAPER_PUBLIC_URL=
--- a/backend/pkg/config/config.go
+++ b/backend/pkg/config/config.go
@@ -190,6 +190,10 @@ type Config struct {
 	ExternalSSLCAPath   string `env:"EXTERNAL_SSL_CA_PATH" envDefault:""`
 	ExternalSSLInsecure bool   `env:"EXTERNAL_SSL_INSECURE" envDefault:"false"`
 
+	// HTTP client timeout in seconds for external API calls (LLM providers, search tools, etc.)
+	// A value of 0 means no timeout (not recommended).
+	HTTPClientTimeout int `env:"HTTP_CLIENT_TIMEOUT" envDefault:"600"`
+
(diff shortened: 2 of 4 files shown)


asdek:
@efe-arv  thank you for the PR!

PR #253, opened 2026-04-10
feat: integrate SAGE persistent memory for cross-session knowledge

l33tdawg:
## Summary

Adds [SAGE](https://github.com/l33tdawg/sage) (Sovereign Agent Governed Experience) as an optional persistent memory layer for cross-session knowledge retention. SAGE memories go through BFT consensus, have confidence scores, and decay over time — giving PentAGI agents institutional knowledge that survives between engagements.

- **SAGE client** (`pkg/sage/client.go`) — Ed25519-authenticated HTTP client with persistent keypair identity
- **SAGE tools** (`sage_recall`, `sage_remember`) — registered in pentester, coder, memorist, and enricher executors
- **Agent prompt awareness** — conditional SAGE instructions in prompt templates (follows existing Graphiti pattern)
- **Docker Compose sidecar** (`docker-compose-sage.yml`) — SAGE node + Ollama for semantic embeddings
- **Configuration** — `SAGE_ENABLED`, `SAGE_URL`, `SAGE_KEY_PATH`, `SAGE_BOT_NAME`, `SAGE_TIMEOUT`

## How it works

When enabled, agents:
1. **Recall** from SAGE at the start of tasks — checking for known vulnerabilities, techniques, and patterns from previous engagements
2. **Store** significant findings after discovery — SQLi payloads, credential dumps, successful techniques
3. **Build institutional knowledge** — each engagement makes future ones more effective

## Test results

Tested against OWASP Juice Shop. SAGE-enhanced run completed **8/8 subtasks** covering 7 vulnerability classes (recon, SQLi, XSS search, XSS reviews, JWT manipulation, IDOR, path traversal). Agents actively recalled from SAGE before testing and stored findings back — **44 SAGE-related messages** observed during the run.

Key behavioral improvements with SAGE enabled:
- Agents check SAGE for prior knowledge before starting each phase
- Known endpoints were targeted immediately instead of discovered through blind enumeration
- Previously untested vectors (JWT, IDOR, path traversal) were prioritized based on SAGE-seeded context
- A dedicated "Store All Findings in SAGE" subtask was auto-created by the agent

Unit tests: 32/32 passing across client, tools, templates, and config packages.

## Usage

\`\`\`bash
# Start with SAGE sidecar
docker compose -f docker-compose.yml -f docker-compose-sage.yml up -d

# Enable in .env
SAGE_ENABLED=true
SAGE_URL=http://sage:8080
SAGE_BOT_NAME=pentagi
\`\`\`

## Files changed

29 files, +2042 lines. Zero changes to core orchestration — SAGE is fully optional and follows the existing Graphiti integration pattern.

## Test plan

- [x] \`go build ./...\` — clean
- [x] \`go vet ./...\` — clean
- [x] Unit tests: 32/32 SAGE tests passing (client, tools, templates, config)
- [x] Integration tests: client health, embed, remember+recall round-trip
- [x] E2E: full pentest against Juice Shop with SAGE-enhanced agents
- [ ] Reviewer runs \`docker compose -f docker-compose.yml -f docker-compose-sage.yml up -d\` and verifies SAGE tools appear in agent tool list


asdek:
hey @l33tdawg 

it looks quite interesting, give us some time to test your product and this integration

l33tdawg:
No worries mate - let me know if you guys have any questions :)

l33tdawg:
Hey @asdek just checking in :)

sickwell:
Hey @l33tdawg and thanks for the interest in PentAGI and for taking the time to put this integration together.

We reviewed the PR in full and also tested the integration in our lab environment using `qwen3.6-35b-a3b` and `kimi k2 (2.5/2.6)`. Our conclusion is fairly straightforward: in its current form, this integration is not something we can justify merging into the product.

### What we observed

In both cases, the model used memory very rarely:
- it almost never saved useful information into memory
- it also almost never came back to retrieve knowledge from it

From our perspective, this is not a flaw in SAGE itself. It is mostly a consequence of the current integration style.

### Why this happens

Right now, the integration is lightweight in the sense that SAGE is presented to the agent as an optional capability. The model can choose from the available tools, but nothing really forces it to rely on SAGE in a consistent way.

In practice, that means the agent tends to prioritize built-in mechanisms — its native reasoning flow, task context, and existing memory-related behavior — over an external memory layer.

So the issue here is not that SAGE does not work. The issue is that, with a soft integration like this, memory usage is too inconsistent to produce reliable benefits.

### Why we are not moving forward with it right now

Based on the tests we ran, the requests we analyzed, and the data that was actually stored and retrieved, this integration does not currently look worthwhile for the product.

To make it genuinely useful, it would require much deeper work around:
- stricter control over when recall/store happens
- tighter integration into the agent workflow
- better evaluation of retrieval quality
- a clearer way to measure whether SAGE actually improves pentest speed or quality

At this stage, that is a much deeper engineering effort than it may seem from the surface.

### Next steps

We would be happy to revisit this direction once we have a dedicated benchmark that can evaluate the actual benefits of SAGE in a more complete and fair way. For us, a successful integration should demonstrate one or more of the following:
- improved pentest quality and speed
- reduced overall LLM usage
- more effective use of weaker models through external memory

Thanks again for the contribution and for sharing this work with us. We hope SAGE continues to evolve, and we would be glad to return to this topic — and potentially collaborate in the future — once we have the right benchmark and evaluation setup in place.

l33tdawg:
Hey @sickwell - really appreciate the depth of review here. Spinning up your lab, testing on two model families, and writing it up properly is a real investment of time, and the feedback is sharp in exactly the way I'd hoped for.

Your read is fair: when memory is offered as just-another-tool, models default to their native flow and the value never shows up. You've also pinned where the work needs to happen - ambient recall and store driven by the workflow, plus measurable proof that it actually helps.

So I'm going to come back with a different shape: a wrapper-style integration where SAGE auto-recalls before each agent step and auto-extracts observations after, plus a small pentest-flavored benchmark (SAGE-on vs SAGE-off, including weaker-model runs since that's where external memory should win hardest). 

Thanks again. Genuinely.
— Dhillon

PR #261, opened 2026-04-15
feat: add TERMINAL_TOOL_TIMEOUT configuration

### LINKED ISSUE #256: Feature request: Expose TERMINAL_TOOL_TIMEOUT as configurable environment variable
## Problem
When running long-running tools like Nikto, the terminal tool times out with:

  terminal tool failed: timeout value is too low, use greater value if you need so: context deadline exceeded

## Environment
- PentAGI version: latest (image created 2026-03-23, revision 3ce09b9)
- Deployment: Docker Compose

- Deployment: Docker Compose

## Expected behavior
A configurable environment variable (e.g. TERMINAL_TOOL_TIMEOUT) in the .env file
to control the terminal tool execution timeout, similar to HTTP_CLIENT_TIMEOUT.

## Suggested variable
TERMINAL_TOOL_TIMEOUT=600  # timeout in seconds, 0 = no timeout

## Workaround
Currently there is no workaround via configuration — the timeout appears to be hardcoded.

mason5052:
## Summary

- add `TERMINAL_TOOL_TIMEOUT` as a configurable environment variable with Docker Compose, config, installer, and docs support
- use the configured timeout as the terminal tool default when a tool call sets `timeout=0`
- add unit coverage for config parsing and terminal timeout normalization

## Problem

Issue #256 reports that long-running terminal commands can time out with no operator-side configuration to change the default behavior.

At the moment, terminal execution falls back to a hardcoded default timeout in the backend. Users running PentAGI through Docker Compose can adjust `HTTP_CLIENT_TIMEOUT`, but there is no equivalent setting for terminal tool execution.

## Solution

Add a new `TERMINAL_TOOL_TIMEOUT` setting and thread it through the relevant execution paths.

- `backend/pkg/config` parses `TERMINAL_TOOL_TIMEOUT` with a default of `600`
- `docker-compose.yml` and `.env.example` expose the variable for Compose-based deployments
- terminal tool constructors now receive the configured default timeout and apply it when a tool call uses `timeout=0`
- `0` is supported as "no server-side default timeout"
- installer server settings now expose the same variable
- docs and tool descriptions were updated so the behavior is visible to users and agents

This keeps explicit tool timeouts intact while giving operators a configuration-level fallback for long-running commands.

## User Impact

Operators can now increase the default terminal execution window for long-running commands without patching the codebase. Compose users can set `TERMINAL_TOOL_TIMEOUT` in `.env`, and installer users can change the same setting from the server settings flow.

Closes #256.

## Test Plan

- [x] `go test ./pkg/config ./pkg/tools/...`
- [x] `go test ./cmd/ftester/... ./cmd/installer/wizard/controller ./cmd/installer/wizard/models ./cmd/installer/wizard/locale`
- [x] `gofmt -w` on changed Go files
- [x] `git diff --check`

## Notes

`go test ./cmd/installer/wizard/...` still hits pre-existing Windows host issues in `cmd/installer/wizard/terminal` because that package expects POSIX commands such as `echo`, `sh`, and `cat` to be available in `PATH`. This PR does not modify that package.


asdek:
@mason5052 thank you for the PR!

asdek:
@mason5052 thanks for the contribution again.

I applied a follow-up commit https://github.com/vxcontrol/pentagi/commit/78e46bd77ac8526859b1f351296634d5453c9ce3 on top of this to adjust the timeout semantics slightly:

- **`TERMINAL_TOOL_TIMEOUT=0` no longer means "no server-side default timeout"** — agents must always have a finite execution bound. instead, `0` (and any negative value) is now clamped to the hard maximum of 3 hours (`maxExplicitExecCommandTimeout`). the default was also raised to `1200 s` (20 min) for backward compatibility with the previous hardcoded value
- the hard ceiling was raised from **20 min to 3 hours**, giving operators more headroom for long-running scans
- `normalizeExecTimeout` was simplified: the operator-configured value (+ 5 s buffer) is now the single dynamic ceiling rather than a fixed constant. explicit agent timeouts within that ceiling pass through unchanged; anything outside falls back to the ceiling
- tool-facing descriptions (agent schema, registry, mock) no longer reference `TERMINAL_TOOL_TIMEOUT` directly - agents only need to know the 3-hour absolute maximum; the actual ceiling is enforced server-side.
- all timeout unit tests were rewritten to cover the new boundary conditions.

the config field stays `int` (accepting negatives) since the env is operator-controlled, with clamping handled in `configuredExecTimeout()` rather than at parse time.
the behavior for negative values is now documented inline.

PR #265, opened 2026-04-15
fix: align compose checks with docker compose

### LINKED ISSUE #254: a problem encountered during installation on Windows 10
Image
Can anyone help me solve this problem? 
I lack experience and don't know how to solve it.

mason5052:
## Summary

- remove the install-time fallback to legacy `docker-compose --version`
- align Docker Compose validation with the actual installer execution path, which already uses `docker compose`
- add unit coverage for the compose version check and update troubleshooting text accordingly

## Problem

Issue #254 reports installation trouble on Windows, and the current installer logic contains a confusing mismatch.

The checker accepts either `docker compose version` or legacy `docker-compose --version`, but the processor later executes only `docker compose ...`. That means the preflight check can succeed in environments where the actual install command will still fail.

## Solution

Make the checker validate the same command path that PentAGI actually uses at runtime.

- `checkDockerComposeVersion()` now treats `docker compose version` as the only source of truth
- a small testable helper was added so the compose validation path can be unit tested without shelling out in tests
- installer troubleshooting text now explains that PentAGI requires the Docker Compose v2 plugin and that legacy `docker-compose` alone is not enough

This keeps the fix narrowly focused on validation/runtime alignment without changing the existing minimum version policy or the compose processor itself.

## User Impact

Users now get an earlier and more accurate failure when the Docker Compose v2 plugin is missing, instead of passing preflight checks and failing later during execution. The installer guidance is also clearer about what needs to be installed.

Closes #254.

## Test Plan

- [x] `go test ./cmd/installer/checker/... -run 'TestCheckDockerComposeVersionWithRunner|TestExtractVersionFromOutput|TestCheckVersionCompatibility'`
- [x] `go test ./cmd/installer/wizard/locale`
- [x] `git diff --check`

## Notes

`go test ./cmd/installer/checker/...` still fails on this Windows host because of a pre-existing `TestCheckFileExistsAndReadable` issue unrelated to this PR. The targeted compose/version tests added in this change pass successfully.

### DIFF
--- a/backend/cmd/installer/checker/helpers.go
+++ b/backend/cmd/installer/checker/helpers.go
@@ -236,14 +236,15 @@ func checkDockerCliVersion() DockerVersion {
 }
 
 func checkDockerComposeVersion() DockerVersion {
-	cmd := exec.Command("docker", "compose", "version")
-	output, err := cmd.Output()
+	return checkDockerComposeVersionWithRunner(func(name string, args ...string) ([]byte, error) {
+		return exec.Command(name, args...).Output()
+	})
+}
+
+func checkDockerComposeVersionWithRunner(run func(name string, args ...string) ([]byte, error)) DockerVersion {
+	output, err := run("docker", "compose", "version")
 	if err != nil {
-		cmd = exec.Command("docker-compose", "--version")
-		output, err = cmd.Output()
-		if err != nil {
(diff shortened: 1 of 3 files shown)


asdek:
@mason5052  thanks a lot! a totally agree with removing `docker-compose` plugin support

Issue #192, opened 2026-03-11
[Enhancement]: Assistant inject in to automation


ThatServerGuy:
### Target Component

Core Services (Frontend UI/Backend API)

### Enhancement Description

Currently Assistant and Automation are relatively independent functions. 

Would like the ability to have them more integrated. 
Use Cases:
1. Prior to any active automation, have the assistant generate structured instructions for the request to feed in to the automation. This should be mapped out as a full plan that the automation function can ingest and spawn the appropriate agents. 
2. When there is an active automation task running, the assistant should be able to inject without interrupting active tasks, i.e. queue a tasks so next break between active tasks the instructions can be included without interruption. This could be to feed new data sources, new instructions, files, etc. 

### Verification

- [x] I have checked that this enhancement hasn't been already proposed
- [x] This enhancement aligns with PentAGI's goal of autonomous penetration testing
- [x] I have considered the security implications of this enhancement
- [x] I have provided clear use cases and benefits

PR #268, opened 2026-04-16
feat: queue input for running automation flows

mason5052:
## Summary

- allow automation flows to accept follow-up input while status is `Running`
- queue submitted input in the backend so it is delivered at the next task boundary instead of requiring an interrupt
- keep `Stop` available in the UI and add queue-focused backend tests

## Problem

Issue #192 asks for tighter assistant and automation integration. A smaller but immediately useful part of that request is use case 2: submit new instructions while automation is already running without interrupting the current task.

Today PentAGI blocks that path in two places.

- the automation UI hides submit while the flow is `Running` and only offers `Stop`
- the backend input path uses an unbuffered handoff, so follow-up instructions are not handled as a reliable queue

## Solution

Implement the queueing path for running automation flows without expanding scope into assistant handoff or file injection yet.

- flow input now uses a bounded FIFO queue with explicit queue-full errors and tests for enqueue, early error, cancel, and stop-drain behavior
- provider switching now happens when queued input is actually processed, so queued messages follow the current flow execution path safely
- the automation form keeps provider changes disabled, but now allows message submission while `Running` and shows `Stop` alongside `Submit`
- queue-focused helper text explains that new instructions are applied at the next task boundary

This intentionally targets use case 2 from #192. Assistant-generated plans, structured handoff, and file or data-source injection can still follow in later work.

## User Impact

Users can add follow-up instructions to a running automation flow without interrupting the active task. The current task continues, the new instruction is queued for the next handoff point, and `Stop` remains available when they do want to cancel execution.

## Test Plan

- [x] `go test ./pkg/controller/...`
- [x] `npx eslint src/features/flows/flow-form.tsx src/features/flows/messages/flow-automation-messages.tsx src/providers/flow-provider.tsx`
- [x] `npm run build`
- [x] `git diff --check`

## Notes

`npm run lint` still fails on this Windows host because of pre-existing unrelated frontend lint errors in files such as `src/components/shared/markdown.tsx`, `src/components/ui/textarea.tsx`, and several existing flow/search components. The files changed in this PR pass the targeted `eslint` command above.

### DIFF
--- a/backend/pkg/controller/flow.go
+++ b/backend/pkg/controller/flow.go
@@ -58,6 +58,7 @@ type flowWorker struct {
 	taskWG  *sync.WaitGroup
 	inputWG *sync.WaitGroup
 	input   chan flowInput
+	inputTO time.Duration
 	discard atomic.Bool
 	flowCtx *FlowContext
 	logger  *logrus.Entry
@@ -589,12 +590,12 @@ func (fw *flowWorker) PutInput(
 	}
 	if err := ctx.Err(); err != nil {
 		close(flin.done)
-		return fmt.Errorf("flow %d input processing timeout: %w", fw.flowCtx.FlowID, err)
+		return fw.inputContextError(err)
 	}
 
 	select {
 	case fw.input <- flin:
-		timer := time.NewTimer(flowInputTimeout)
+		timer := time.NewTimer(fw.flowInputTimeout())
 		defer timer.Stop()
 
 		select {
@@ -605,7 +606,7 @@ func (fw *flowWorker) PutInput(
 		case <-fw.ctx.Done():
 			return fmt.Errorf("flow %d stopped: %w", fw.flowCtx.FlowID, fw.ctx.Err())
 		case <-ctx.Done():
-			return fmt.Errorf("flow %d input processing timeout: %w", fw.flowCtx.FlowID, ctx.Err())
+			return fw.inputContextError(ctx.Err())
 		}
 	default:
 		if err := fw.ctx.Err(); err != nil {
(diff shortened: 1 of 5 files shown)


asdek:
@mason5052 thanks again - you've been a consistent and high-quality contributor and this PR is no exception in terms of engineering care.

after reviewing the approach carefully I'm going to decline merging it in its current shape, and I want to explain exactly why so the reasoning is clear

**the core concern: in-memory queue fragility**

the buffered-channel queue solves the immediate "don't block the caller" problem, but it introduces a class of issues we'd be inheriting permanently:

- **no persistence.** a restart, crash, or container restart silently drops everything in the queue. users who submitted follow-up instructions would never know they were lost, and we'd have no way to surface that.
- **no visibility.** there's nothing in the UI, the database, or the GraphQL schema that lets a user see what's waiting in the queue. they submit and… hope. this is especially problematic for automation flows that can run for tens of minutes.
- **no management.** there is no way to cancel, reorder, or inspect queued items. once something is in the channel it either gets processed or silently dropped on stop/drain. that's a regression in user control compared to the current hard-stop model.

**the task-boundary model creates runaway flows**

the deeper architectural issue is what happens at the "next task boundary". PentAGI's current model is `flow -> tasks -> subtasks -> actions` with deterministic lifetime. when queued input is consumed, the flow will spawn a new top-level task (or equivalent). without a hard bound on the queue depth and without the user being able to see the outstanding task graph, the flow becomes open-ended in a way that's very hard to reason about. users would have no reliable signal for "this flow is done."

thank you for tackling this - the problem itself is real and worth solving properly.

asdek:
@mason5052 I've tried to implement this algorithm here https://github.com/vxcontrol/pentagi/commit/5067e8f5a44456984f8b3401d47e8aff4722189b
what do you think about that?

mason5052:
Thanks for the detailed review and for spelling out the concerns so clearly. I agree the in-memory queue version introduces too much hidden state around persistence, visibility, and flow lifetime, so closing this version makes sense.

I also looked through \5067e8f5a44456984f8b3401d47e8aff4722189b\, and I think the flow-management-tool direction is much more aligned with PentAGI's task model. I'm happy to treat this PR as superseded and follow up later with tests or docs around that approach if that would be useful.

Issue #191, opened 2026-03-11
[Enhancement]: Ability to configure LLM's, Search providers, Third party APIs, etc from web console


ThatServerGuy:
### Target Component

Core Services (Frontend UI/Backend API)

### Enhancement Description

Currently you cannot configure any settings in the web UI and only through the yaml or the installer script. 
Once the environment is stood up would be great to have the ability to configure LLM's, Search providers, Third party APIs, etc from web console

### Verification

- [x] I have checked that this enhancement hasn't been already proposed
- [x] This enhancement aligns with PentAGI's goal of autonomous penetration testing
- [x] I have considered the security implications of this enhancement
- [x] I have provided clear use cases and benefits

PR #270, opened 2026-04-16
docs: clarify current web settings coverage

mason5052:
## Summary
- clarify which settings areas are already managed in the web UI
- document which configuration still lives in environment variables, compose files, or mounted config files

## Problem
Issue #191 asks for configuration from the web console, but the current docs do not clearly separate shipped web settings from configuration that is still server-managed.

That makes it easy to overestimate current UI coverage, especially around search providers, third-party integrations, and LLM credentials or endpoint settings.

## Solution
- add a short `Current Web Settings Coverage` section to `README.md`
- add a matching clarification near the top of `backend/docs/config.md`
- explicitly distinguish provider profiles, prompts, and API tokens from server-managed credentials, search providers, third-party integrations, and MCP server setup

## User Impact
Users get a more accurate picture of what can already be configured after startup and what still has to be configured on the server.

This also makes #191 easier to treat as a follow-up on the remaining gap instead of a blind reimplementation of features that already exist.

## Test Plan
- re-checked each capability claim against the current frontend settings pages, GraphQL schema and resolvers, and config package
- ran `git diff --check`

References #191.

### DIFF
--- a/README.md
+++ b/README.md
@@ -625,6 +625,24 @@ The installer will:
 5. **Security Hardening**: Generate secure credentials and configure SSL certificates
 6. **Deployment**: Start PentAGI with docker-compose
 
+### Current Web Settings Coverage
+
+The PentAGI web console already manages several settings areas after the server is up and running:
+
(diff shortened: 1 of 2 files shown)


asdek:
@mason5052 thanks!

Issue #193, opened 2026-03-11
[Enhancement]: File uploads for Flows


ThatServerGuy:
### Target Component

Core Services (Frontend UI/Backend API)

### Enhancement Description

Add file upload support scoped to individual workflows, with files stored in a shared location accessible by everything the workflow spawns — the Assistant, Automation, and the container.
Requirements:
	∙	Upload interface added to the workflow UI
	∙	Shared storage mounted and accessible across all spawned functions (Assistant, Automation, Docker container)
	∙	Files isolated and scoped to their parent workflow
	∙	On workflow deletion, prompt user to archive or permanently delete associated files before removal is completed​​​​​​​​​​​​​​​​

### Verification

- [x] I have checked that this enhancement hasn't been already proposed
- [x] This enhancement aligns with PentAGI's goal of autonomous penetration testing
- [x] I have considered the security implications of this enhancement
- [x] I have provided clear use cases and benefits

vchaplygim:
Closing as completed (since v2.1.0).

Delivered:
- Files tab in the workflow UI;
- shared storage for Assistant, Automation, and the container (/work/uploads, /work/resources);
- per-flow isolation (flow-{id}-data/).

On deletion: the flow is soft-deleted (deleted_at), the container is removed on teardown, and the host file directory is retained on purpose so artifacts remain available for inspection. This matches soft-delete semantics and covers post-hoc investigation without an archive/purge prompt at delete time. While the flow still exists, files can be removed manually from the Files tab.

PR #272, opened 2026-04-19
feat: add flow-scoped file uploads

mason5052:
## Summary
This PR adds a small v1 slice of flow-scoped file uploads.

## Problem
Issue #193 asks for files that can be attached to a flow and then shared across the flow runtime. PentAGI already gives each flow a shared `/work` directory, but there was no UI or API path for users to place files into that workspace.

## Solution
This PR adds a minimal flow files vertical:
- a backend REST surface to list, upload, and download user-uploaded files for a flow
- storage under the existing flow workspace at `/work/uploads`
- a new `Files` tab in the flow UI with upload, refresh, filter, download, and copy-path actions
- focused backend tests for filename sanitization and filesystem listing behavior

This intentionally does not try to solve archive/export/delete lifecycle yet. It keeps the change small and reviewable while making uploaded files immediately available to automation, assistants, and container commands in the same flow.

## User Impact
Users can now upload files directly from a flow page and reference them inside the flow runtime at `/work/uploads/<filename>`. That makes it practical to seed a flow with wordlists, configs, notes, or other local inputs without adding a broader asset platform first.

## Test Plan
- `go test ./pkg/server/services -run FlowFile`
- `npx eslint src/features/flows/flow-tabs.tsx src/features/flows/files/flow-files.tsx`
- `npx tsc --noEmit`
- `git diff --check`

Refs #193

### DIFF
--- a/backend/pkg/server/response/errors.go
+++ b/backend/pkg/server/response/errors.go
@@ -113,7 +113,8 @@ var ErrFlowsInvalidData = NewHttpError(500, "Flows.InvalidData", "invalid flow d
 
 var ErrFlowFilesInvalidRequest = NewHttpError(400, "FlowFiles.InvalidRequest", "invalid flow file request data")
 var ErrFlowFilesNotFound = NewHttpError(404, "FlowFiles.NotFound", "flow file not found")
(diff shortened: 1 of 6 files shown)


asdek:
@mason5052 thank you for the PR! we'll added files control in the separated tab with file browser

Issue #235, opened 2026-04-05
RFC: Cryptographic evidence chain for PentAGI pentest operations


tomjwxf:
## Problem

PentAGI runs automated penetration testing with AI agents. Security assessments produce findings that need to be demonstrably authentic and tamper-proof — a modified pentest report could hide vulnerabilities or fabricate findings. Currently, there is no cryptographic evidence chain proving what tools were run, what was discovered, and that the results are unmodified.

## Proposal

Integrate Ed25519 receipt signing into PentAGI's tool execution pipeline. Each pentest tool invocation would produce a signed receipt creating a verifiable evidence chain:

```
Recon (receipt_001) → Port Scan (receipt_002) → Exploit Attempt (receipt_003) → Report (receipt_004)
```

This transforms pentest results from "trust me" reports into cryptographically verifiable evidence packages.

## Reference

[protect-mcp](https://github.com/scopeblind/scopeblind-gateway) (MIT, v0.5.3). Receipt format: [IETF Internet-Draft](https://datatracker.ietf.org/doc/draft-farley-acta-signed-receipts/). The receipt chain/DAG visualization is built-in via `npx protect-mcp trace <id>`.

Happy to discuss and contribute.

sirozha:
Not planned: no demand for signed receipts in the product. Closing — comment here if that changes.

PR #277, opened 2026-04-22
docs: add evidence chain RFC

mason5052:
## Summary
- add an RFC for cryptographic evidence-chain receipts in PentAGI operations
- outline signed receipt fields, Ed25519 signing model, pipeline integration points, verification workflow, and open design questions
- link the RFC from the flow execution documentation

## Validation
- `git diff --check`
- verified the flow execution link target exists

Refs #235

### DIFF
--- a/backend/docs/evidence_chain_rfc.md
+++ b/backend/docs/evidence_chain_rfc.md
@@ -1,0 +1,118 @@
+# RFC: Cryptographic Evidence Chain for PentAGI Operations
+
+Issue [#235](https://github.com/vxcontrol/pentagi/issues/235) proposes signed evidence receipts for PentAGI pentest operations. This RFC captures a possible design direction without committing the project to an implementation.
+
+## Goals
+
+- Make reports verifiable against the tool actions, artifacts, and agent decisions that produced them.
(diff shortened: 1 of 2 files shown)


asdek:
hey @mason5052 

thanks for the PR!

the topic of creating verified kill chains is definitely relevant for various bug bounty programs, but i don't think this particular approach makes sense for an on-premises product. it would only be useful if there are multiple departments within an organization where one passes information to another, and they completely trust the product itself (that its code can't somehow falsify data) and that the department using PentAGI won't have access to the private keys. there are quite a few conditions in this reasoning that are unlikely to align in real life. this mechanism could work for cloud products that keep their private key secret and where the receiving party can verify the kill chain by accessing the service's public key and consider all received content trusted.

overall, we can merge this PR - at minimum the proposed implementation content makes sense, but the actual implementation would require long discussions not only about the implementation itself but also about the user scenario where this functionality would be useful, and i don't have an answer yet as to why this might be needed.

i understand why it's necessary in principle to verify each artifact that gets added to a report, but using cryptography for this seems like quite a significant complication for local product usage.

Issue #68, opened 2025-10-25
[Enhancement]: support for multiable osint frameworks that agents can use


s-b-repo:
### Target Component

Security Tools Integration

### Enhancement Description

app.flare.io and also ransomware.live

### Verification

- [ ] I have checked that this enhancement hasn't been already proposed
- [ ] This enhancement aligns with PentAGI's goal of autonomous penetration testing
- [ ] I have considered the security implications of this enhancement
- [ ] I have provided clear use cases and benefits

asdek:
Hello @s-b-repo 

I looked into both solutions and honestly need some clarification on how they would be useful:

**ransomware.live**
Looking at their [API](https://api-pro.ransomware.live/docs#/), which specific endpoints do you think would be valuable for agents? When would they use this data and what benefit would it provide?

**flare.io** 
This service does perimeter scanning and feeds analysis results into clients' defensive tools. How do you envision this integration working? What output would you expect? 

Could you describe an end-to-end scenario? For instance:
- Agent detects new domain/new service -> queries Flare API -> gets threat intel -> takes action X
- Or something different?

sirozha:
Not planned: no end-to-end scenario was provided. Closing — comment here if that changes.

PR #278, opened 2026-04-22
docs: add OSINT integration scenarios

mason5052:
## Summary
- add provider-neutral OSINT integration scenarios for ransomware.live and Flare
- answer when agents should query these providers, which data categories are useful, and what outputs should feed reports
- document guardrails so OSINT enrichment does not expand pentest scope or encourage credential misuse
- link the guide from the external functions configuration docs

## Validation
- `git diff --check`
- verified the config docs link target exists
- checked current ransomware.live and Flare API docs while drafting endpoint categories and scenarios

Refs #68


asdek:
hey @mason5052  thanks for the PR, i've moved the documents from this PR and #277 into a separate examples/proposals folder to avoid confusing users and keep the documentation current, you can see the changes here
https://github.com/vxcontrol/pentagi/commit/47de4e44e639f2b8154aaff7f8ab715dc3d6cfb3

PR #279, opened 2026-04-22
feat: add evidence receipt hash chain prototype

mason5052:
## Summary
- add a disabled-by-default Evidence Receipts v1 prototype for toolcall completion
- write hash-chain JSONL receipts under `<DATA_DIR>/flow-<flow_id>/evidence/receipts.jsonl`
- include toolcall provenance metadata plus argument/result hashes without storing raw args or raw results
- document v1 as hash-chain-only; Ed25519 signing and report bundle export are deferred

## Validation
- `go test ./pkg/tools/...`
- `go test ./pkg/config/...`
- `git diff --cached --check`

Refs #235


asdek:
hey @mason5052 
thank you for the PR and for the effort you put into this, the hash-chain receipts approach is well thought out.

however, this doesn't align with our current development direction. the provenance/audit functionality you're proposing is already fully covered by our ongoing Graphiti integration, so we won't need equivalent changes on the backend side for this.

I'll close this PR for now, but feel free to reach out if you'd like to discuss how this could tie into the Graphiti work instead.

PR #292, opened 2026-04-28
docs: explain assistant flow management for active flows

mason5052:
## Summary
- document how the Assistant view can manage an active flow without restarting the automation
- explain the explicit control path for checking status, stopping work, submitting follow-up input, and patching planned subtasks
- keep the guidance in the README first-use section where users are most likely to look for this workflow

## Problem
Issue #192 shows that users want a safe way to influence an in-progress automation. The implementation direction has shifted away from an in-memory queue and toward explicit assistant-side flow management tools, but that workflow is still hard to discover from the main README.

## Solution
Add a short user-facing section to the first-use guide that explains when to open the Assistant view for the same flow and what kinds of changes it is appropriate to make there.

## User Impact
Users get a clearer path for steering an active flow without assuming PentAGI uses a hidden background queue. This should reduce confusion around how to redirect or clarify an automation run after it has already started.

## Test Plan
- verify the new README section matches the current assistant flow-management behavior documented in `backend/docs/flow_execution.md`
- verify the wording does not promise hidden queueing or direct mid-command interruption
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1017,6 +1017,15 @@ Once the flow has enough results, use the **Report** menu on the flow page to:
 - download the report as Markdown
 - download the report as PDF
 
+### 4. Use the Assistant view to steer an active flow
+
+Each flow also includes an **Assistant** view for interactive guidance. This is useful when the autonomous run uncovers something that needs human direction instead of a hard restart.
+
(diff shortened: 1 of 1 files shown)


asdek:
thanks!

PR #294, opened 2026-04-29
docs: clarify web UI account setup

### LINKED ISSUE #289: Needs clear attention on how to make an account for the UI
How can i sign up for the account for the WEB UI

mason5052:
Summary
- Adds a short Web UI Accounts section to the Quick Start flow.
- Clarifies that PentAGI does not provide public self-service sign-up from the login page.
- Documents the default local admin account, first-login password change expectation, installer reset path, and the authenticated Users REST API path for multi-user setups.

Problem
New users can see the Web UI login screen but the README only mentioned the default credentials in one line. That made it unclear whether they should sign up from the UI, use the seeded admin account, or manage additional users elsewhere.

Solution
I added a focused README note beside the Web UI access instructions, grounded in the current seeded admin account, password reset maintenance screen, and `/api/v1/users/` REST API.

User Impact
First-time operators get a clearer path into the UI and a more accurate expectation for multi-user setup without implying that public registration is available.

Test Plan
- Verified the default admin account against the initial database migration.
- Verified the installer has a reset-admin-password maintenance flow.
- Verified the Users REST API routes exist in the server router.
- Ran `git diff --check`.

Closes #289

### DIFF
--- a/README.md
+++ b/README.md
@@ -762,6 +762,17 @@ docker compose up -d
 
 Visit [localhost:8443](https://localhost:8443) to access PentAGI Web UI (default is `admin@pentagi.com` / `admin`)
 
+#### Web UI Accounts
+
+PentAGI does not expose public self-service sign-up from the login page. A fresh installation creates the default local administrator account:
+
+- **Email**: `admin@pentagi.com`
+- **Password**: `admin`
+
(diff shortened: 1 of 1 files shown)


asdek:
thanks!

Issue #313, opened 2026-05-13
[Bug]: Tool call calling issue in combination with llama.cpp backend


SebastianGode:
### Affected Component

AI Agents (Researcher/Developer/...)

### Describe the bug

When trying to run a simple test a few calls work and afterwards there's no response anymore in the UI to new questions:

<img width="1670" height="1037" alt="Image" src="https://github.com/user-attachments/assets/6a650360-737d-4e25-b414-1caab332a3f8" />

Inside the log I get the following:
`pentagi     | time="2026-05-13T11:26:02Z" level=error msg="failed to perform assistant agent chain" error="failed to call agent chain: max retries reached, 3: API returned unexpected status code: 500: litellm.InternalServerError: InternalServerError: Custom_openaiException - Failed to parse tool call arguments as JSON: [json.exception.parse_error.101] parse error at line 1, column 131: syntax error while parsing value - unexpected end of input; expected '[', '{', or a literal. Received Model Group=qwen3.6-35b\nAvailable Model Group Fallbacks=None\nAPI returned unexpected status code: 500: litellm.InternalServerError: InternalServerError: Custom_openaiException - Failed to parse tool call arguments as JSON: [json.exception.parse_error.101] parse error at line 1, column 131: syntax error while parsing value - unexpected end of input; expected '[', '{', or a literal. Received Model Group=qwen3.6-35b\nAvailable Model Group Fallbacks=None\nAPI returned unexpected status code: 500: litellm.InternalServerError: InternalServerError: Custom_openaiException - Failed to parse tool call arguments as JSON: [json.exception.parse_error.101] parse error at line 1, column 131: syntax error while parsing value - unexpected end of input; expected '[', '{', or a literal. Received Model Group=qwen3.6-35b\nAvailable Model Group Fallbacks=None\nfailed to perform caller reflector: failed to call agent chain: max retries reached, 3: API returned unexpected status code: 429: No deployments available for selected model, Try again in 30 seconds.`

The llama.cpp servers also show the following: 
```
srv  log_server_r: done request: POST /v1/chat/completions 10.0.0.40 500
srv    operator(): got exception: {"error":{"code":500,"message":"Failed to parse tool call arguments as JSON: [json.exception.parse_error.101] parse error at line 1, column 131: syntax error while parsing value - unexpected end of input; expected '[', '{', or a literal","type":"server_error"}}
```

### Steps to Reproduce

1. Access PentAGI UI
2. Open a new Flow
3. Do a simple task

[...]

PR #330, opened 2026-06-04
docs: add tool-call parser troubleshooting for custom LLM backends

mason5052:
## Summary

Add a troubleshooting subsection under **Custom LLM Provider Configuration** explaining why tool-call (function-call) parser problems on self-hosted OpenAI-compatible backends (llama.cpp / SGLang / vLLM, often behind LiteLLM) cause stalled flows, and how to diagnose them. Docs only.

## Problem

Issue #313 reported that flows stop responding after a few steps when running a custom backend configured through `LLM_SERVER_*` (LiteLLM in front of llama.cpp serving `qwen3.6-35b`). The logs showed:

```
Failed to parse tool call arguments as JSON: [json.exception.parse_error.101]
parse error at line 1, column 131: syntax error while parsing value -
unexpected end of input
```

surfaced through LiteLLM as an HTTP 500, followed by cascading retries and a 429. The maintainer confirmed the stall was fixed in the latest build by sanitizing malformed function-call arguments, and that the root cause was the model side returning corrupted tool-call arguments.

There is currently no documentation that explains this class of failure, even though it is a common pitfall with self-hosted backends and is closely related to the image-chooser failure (a flow's first action is an LLM tool call to pick the container image).

## Solution

Add a `#### Troubleshooting: tool-call (function-call) parser errors` subsection right after the Custom LLM Provider Configuration content, covering:

- Custom OpenAI-compatible backends must return valid tool-call JSON; llama.cpp, SGLang, and vLLM usually require a specific tool-call parser and a matching chat template, and not every setup produces valid tool calls out of the box (compatibility depends on the backend, not PentAGI alone).
- Symptoms: `Failed to parse tool call arguments as JSON`, a flow that stalls after a few steps, looping tool calls, the start-of-flow `failed to select primary docker image via llm call` error, and unexpected backend 5xx/4xx responses.
- How to investigate: check both PentAGI and backend/proxy logs, validate the provider with `ctester` before a full flow, confirm the parser/chat template match the model, and update PentAGI (recent builds sanitize malformed function-call arguments).

The new content links only to the existing [Testing LLM Agents](https://github.com/vxcontrol/pentagi#testing-llm-agents) section and references the image-chooser error in prose (no new anchor), so it stands on its own against `main`.

## User Impact

- Users on self-hosted/llama.cpp/SGLang/vLLM backends get a clear explanation of the tool-call parser failure mode and a concrete diagnosis path, instead of an opaque `Failed to parse tool call arguments as JSON` stall.
- Points users at `ctester` for pre-flight validation and at the update that sanitizes malformed arguments.
- No behavior change.

## Test Plan

- [x] `git diff --check` clean.
- [x] Docs-only diff: `README.md` (+20 lines). No tool-call parser code, provider runtime, schema, migration, or config-default changes.
- [x] Verified the referenced error string `failed to select primary docker image via llm call` exists in `backend/pkg/providers/providers.go` on `main`.
- [x] Verified `LLM_SERVER_URL` / `LLM_SERVER_KEY` / `LLM_SERVER_MODEL` / `LLM_SERVER_PROVIDER` exist in `.env.example`.
- [x] Verified the `ctester` utility exists and tests tool-calling agent types, and that the `#testing-llm-agents` anchor resolves.
- [x] Placed away from the README regions touched by open PRs #325 and #327 to avoid conflicts.
- [x] No unrelated files included.

Refs #313

### DIFF
--- a/README.md
+++ b/README.md
@@ -1392,6 +1392,26 @@ The `LLM_SERVER_PRESERVE_REASONING` setting controls whether reasoning content i
 
 This setting is required by some LLM providers (e.g., Moonshot) that return errors like "thinking is enabled but reasoning_content is missing in assistant tool call message" when reasoning content is not included in multi-turn conversations. Enable this setting if your provider requires reasoning content to be preserved.
 
+#### Troubleshooting: tool-call (function-call) parser errors
+
+PentAGI drives its agents with tool calls (also called function calls), so any custom OpenAI-compatible backend configured through `LLM_SERVER_*` must return valid tool-call JSON in the format the OpenAI Chat Completions API defines. When the backend emits malformed, truncated, or non-conforming tool-call arguments, the agent chain cannot continue.
+
+Self-hosted engines such as llama.cpp, SGLang, and vLLM usually require a specific tool-call parser and a matching chat template to produce correct tool-call output. If the parser is missing or mismatched for the model you are serving, tool-call arguments can come back corrupted. Compatibility therefore depends on the backend's tool-call/function-call behavior and configuration, not on PentAGI alone; not every llama.cpp or SGLang setup produces valid tool calls out of the box.
+
+Typical symptoms:
+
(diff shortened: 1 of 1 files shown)


asdek:
hey @mason5052 

thank you for the PR!

Issue #335, opened 2026-06-05
[Bug]: xsstrike: error: unrecognized arguments: -c -o /dev/null


Archerkong:
### Affected Component

Security Tools Integration

### Describe the bug

<img width="688" height="195" alt="Image" src="https://github.com/user-attachments/assets/2c43c373-d889-4b34-87b0-484936eecab2" />

### Steps to Reproduce

none

### System Configuration

docker

### Logs and Artifacts

Testing XSS on http://10.1.194.224/phpinfo.php?item=http://example.com/shell.php
usage: xsstrike [-h] [-u TARGET] [--data PARAMDATA] [-e ENCODE] [--fuzzer]
                [--update] [--timeout TIMEOUT] [--proxy] [--crawl] [--json]
                [--path] [--seeds ARGS_SEEDS] [-f ARGS_FILE] [-l LEVEL]
                [--headers [ADD_HEADERS]] [-t THREADCOUNT] [-d DELAY] [--skip]
                [--skip-dom] [--blind]
                [--console-log-level {DEBUG,INFO,RUN,GOOD,WARNING,ERROR,CRITICAL,VULN}]
                [--file-log-level {DEBUG,INFO,RUN,GOOD,WARNING,ERROR,CRITICAL,VULN}]
                [--log-file LOG_FILE] [-n PAYLOAD_COUNT]
xsstrike: error: unrecognized arguments: -c -o /dev/null

### Verification

- [ ] I have checked that this issue hasn't been already reported
- [ ] I have provided all relevant configuration files (with sensitive data removed)
- [ ] I have included relevant logs and error messages
- [x] I am running the latest version of PentAGI

sirozha:
Already available in the current version — closing. Comment here if it does not work that way for you.

PR #343, opened 2026-06-06
fix: avoid unsupported XSStrike flags in pentester prompt

mason5052:
## Summary

Add a narrow Pentester prompt guardrail for XSStrike command construction and cover it with a template-rendering regression test.

## Problem

Issue #335 reports PentAGI executing `xsstrike` with unsupported `-c` and `-o /dev/null` arguments. Investigation did not find a hardcoded XSStrike command or `-c -o /dev/null` pattern in repo prompts, schemas, config, examples, or docs. The issue evidence shows the installed XSStrike help output rejects those flags, so the likely failure mode is model-generated CLI flag reuse during pentest command composition.

## Solution

Update `backend/pkg/templates/prompts/pentester.tmpl` with a small CLI argument protocol that tells the Pentester agent to verify tool-specific flags, avoid copying flags across tools, and specifically avoid `xsstrike -c` or `xsstrike -o` unless the installed `xsstrike --help` documents them.

Add `TestPentesterPromptXSStrikeArgumentGuidance` to ensure the rendered Pentester prompt keeps this guardrail in place.

## User Impact

Users running XSS testing flows should be less likely to hit the reported XSStrike argument error. This does not change runtime tool execution, Docker image contents, schemas, frontend behavior, or unrelated tools.

## Test Plan

- `go test ./pkg/templates` - passed in a Go-enabled verification environment.
- `git diff --check` - passed.
- `rg -n -i "xsstrike|xss strike|xssstrike" backend frontend examples README.md --glob '!backend/pkg/tools/testdata/sploitus_result_nginx.json'` - only the new guardrail and regression test mention XSStrike.
- `rg -n -- "xsstrike -c|xsstrike -o|-o /dev/null|unrecognized arguments" backend frontend examples README.md` - only the new negative guidance/test references `xsstrike -c` and `xsstrike -o`; no stale executable example remains.
- `docker image ls vxcontrol/kali-linux` / `xsstrike --help` - not run locally because Docker is not installed in this environment; XSStrike CLI behavior was verified from the Issue #335 evidence.

Refs #335

## Update: guardrail hardening

- The CLI argument protocol now explicitly forbids `xsstrike -o /dev/null` (the exact flag combination reported in #335) and documents the correct alternative for saving, reducing, or discarding output (shell redirection such as `> /dev/null`), addressing the root cause: the model substituting an unsupported output flag.
- `TestPentesterPromptXSStrikeArgumentGuidance` was extended to also assert the `xsstrike -o /dev/null` negative guidance and the shell-redirection alternative.

Validation limitation: this is a prompt-level guardrail; it reduces but cannot guarantee model behavior. Runtime reproduction of #335 against a live PentAGI + LLM was not performed in this environment, so efficacy is validated structurally (the prompt now names the exact failing flags and the supported alternative) and via the template regression test, not by live execution.

### DIFF
--- a/backend/pkg/templates/prompts/pentester.tmpl
+++ b/backend/pkg/templates/prompts/pentester.tmpl
@@ -323,6 +323,12 @@ Check tool availability with 'which [tool]' before use. Install missing tools if
 {{end}}
 </usage_notes>
 
+<cli_argument_protocol>
+- Verify command-specific flags with `[tool] -h` or `[tool] --help` before first use when the exact syntax is uncertain.
+- Do not copy flags between different tools. For XSStrike, do not use `xsstrike -c` or `xsstrike -o` unless the installed `xsstrike --help` explicitly documents those options.
+- If output needs to be saved or reduced, prefer shell redirection or the tool's documented logging option instead of inventing unsupported output flags.
+</cli_argument_protocol>
+
 <msf_workflow_protocol>
 Standalone (recommended): All operations in one command
 `msfconsole -q -x "use exploit/...; set LPORT [allocated]; exploit; sleep 20; sessions -l; sessions -i 1 -c 'sysinfo'; exit"`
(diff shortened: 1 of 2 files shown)


asdek:
thank you!

Issue #337, opened 2026-06-05
[Bug]: PentAGI Container Escape via Prompt Injection


ez-lbz:
### Affected Component

AI Agents (Researcher/Developer/...)

### Describe the bug

# PentAGI Container Escape via Prompt Injection

## Summary

PentAGI mounts the host Docker socket (`/var/run/docker.sock`) into agent sandbox containers when deployed via Docker Compose, allowing a prompt-injected AI agent to escape its sandbox and execute arbitrary commands on the **host** through the Docker API.

---

## Root Cause

In `backend/pkg/docker/client.go:282-284`, when `DOCKER_INSIDE` is true, every per-flow sandbox container is created with the host Docker socket bind-mounted:

```go
if dc.inside {
    hostConfig.Binds = append(hostConfig.Binds,
        fmt.Sprintf("%s:%s", dc.socket, defaultDockerSocketPath))
}
```

`dc.inside` is controlled by the `DOCKER_INSIDE` environment variable, which defaults to `true` in the shipped `docker-compose.yml`. The `defaultDockerSocketPath` is `/var/run/docker.sock`.

---

## Impact

With host Docker socket access, an attacker can:

- **Host compromise**: `docker run --privileged -v /:/host ...` mounts the host root filesystem into a new container, providing full read/write access to the host.
- **Lateral movement**: Access to all containers, images, volumes, and networks on the host.

### Steps to Reproduce

none

### System Configuration

none

### Verification

- [ ] I have checked that this issue hasn't been already reported
- [ ] I have provided all relevant configuration files (with sensitive data removed)
- [ ] I have included relevant logs and error messages
- [ ] I am running the latest version of PentAGI

ez-lbz:
Sandbox Container: pentagi-terminal-3
 │
 ├─ /var/run/docker.sock (Successfully mounted, srw-rw----)
 │
 ├─ POST /containers/create  → Privileged Container with Volume Mount (-v /:/host)
 │        └─ 201 Created
 ├─ POST /containers/{id}/start → 204 No Content
 ├─ POST /containers/{id}/wait  → StatusCode=0
 ├─ GET  /containers/{id}/logs  → Host filesystem data exfiltration (Leakage)
 └─ DELETE /containers/{id}     → Footprint cleanup / Clearing tracks

sirozha:
[...]

PR #355, opened 2026-06-24
fix(sandbox): harden sandbox containers against privilege escalation

manus-pi:
## Summary

Four targeted container hardening changes for issue #337.

**Important clarification (flagged by code review):** `no-new-privileges` alone does *not* close the Docker socket escape vector — an attacker with socket access can ask the *host daemon* to launch a privileged container regardless of this flag. The changes here are genuine defense-in-depth, but the definitive fix for #337 remains `DOCKER_INSIDE=false` or fronting the socket with a least-privilege proxy. Issue #337 should stay open until that is addressed.

### `backend/pkg/docker/client.go` — RunContainer hardening

- **`no-new-privileges:true`**: prevents setuid/setgid and file-capability escalation inside the container
- **`PidsLimit=2048`**: fork-bomb / resource exhaustion guard
- **Startup warning**: logrus.Warn when DOCKER_INSIDE=true so operators see the socket-escape risk at boot
- **Comment corrected**: no longer implies no-new-privileges mitigates the #337 socket escape (it doesn't)

### `backend/pkg/tools/tools.go` — CapDrop ALL

Drop all Linux capabilities, add back only NET_RAW + optionally NET_ADMIN. Removes ~14 unnecessary caps (SYS_CHROOT, AUDIT_WRITE, MKNOD, etc.) from the default Docker set.

### `.env.example`

- Flip DOCKER_INSIDE default from true → false (aligns with config.go envDefault:"false")
- Expand comment with socket-escape risk, DinD guidance, docker-socket-proxy link

## What is NOT addressed (why #337 stays open)

The root cause — a process in the sandbox can POST to the Docker socket to launch a privileged container with the host filesystem mounted — is not addressed by these changes. Definitive mitigations are: DOCKER_INSIDE=false, a socket proxy, or rootless/sysbox DinD. This PR does the hardening safe to ship immediately while that architectural question is resolved.

## Testing

- `go build ./pkg/docker/... ./pkg/tools/...` — clean
- `go test ./pkg/docker/... ./pkg/tools/...` — all pass

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -213,7 +213,14 @@ DOCKER_TLS_VERIFY=
 DOCKER_CERT_PATH=
 
 ## Docker settings inside primary terminal container
-DOCKER_INSIDE=true # enable to use docker socket
+# SECURITY NOTE: DOCKER_INSIDE=true mounts the host Docker socket into every
+# sandbox container so agents can launch sub-containers (Docker-in-Docker).
+# This grants any process inside the sandbox unrestricted access to the Docker
+# API, which can be exploited via prompt injection to escape the sandbox and
+# reach the host (issue #337).  Only enable this when DinD is required and you
+# understand the risk.  Consider using a least-privilege socket proxy such as
+# https://github.com/Tecnativa/docker-socket-proxy instead of the raw socket.
+DOCKER_INSIDE=false # enable to use docker socket (see security note above; default false)
 DOCKER_NET_ADMIN=true # enable to use net_admin capability
 DOCKER_SOCKET=/var/run/docker.sock # path on host machine
 DOCKER_NETWORK=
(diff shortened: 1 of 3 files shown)


asdek:
hey @manus-pi 
thank you for the PR!

psyray:
@manus-pi 
Thanks for this PR
But it remove the possibility in a flow to create a TUN connection
It was working before this hardening.
For the moment the only way I've found to create /dev/net/tun is by running on the host the following command
```
docker exec --privileged -u root pentagi-terminal-112 sh -c "mkdir -p /dev/net && [ -c /dev/net/tun ] || mknod /dev/net/tun c 10 200; chmod 600 /dev/net/tun"
```

Before this PR, the command could be run inside because of the MKNOD cap.
I fully understand your hardening, and the drop of dangerous capabilities.
But this remove the possibility to keep the flow autonomous

Do you have an idea on how to keep docker config hardened but establish openvpn tunnel ?
Let the user the option, when needed, to bind /dev/net/tun device from host ?
It keeps docker hardened and it could be attached to the DOCKER_NET_ADMIN option (it needs tun module to be loaded on the host).

My use case for the moment is HackTheBox VPN, but in some engagement it could be needed too.

PR #356, opened 2026-06-24
fix(sandbox): add docker-socket-proxy to restrict DinD API access (issue #337)

manus-pi:
Root cause fix for #337. When DOCKER_INSIDE=true, sandbox containers received the raw host /var/run/docker.sock, allowing full host escape. This fronts the socket with Tecnativa/docker-socket-proxy — an allowlist proxy that permits only what agents need for DinD while blocking privileged container creation, exec into siblings, secrets, and system ops. Complements PR #355.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -213,9 +213,32 @@ DOCKER_TLS_VERIFY=
 DOCKER_CERT_PATH=
 
 ## Docker settings inside primary terminal container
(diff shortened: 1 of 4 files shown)


sirozha:
Thanks for the patch. We are not taking this: the proxy filters by path and method, so a privileged container can still be created, and it puts the sandbox on the network that hosts the database. Closing this PR.

PR #369, opened 2026-07-09
fix: stream ZIP downloads directly to response to prevent OOM

### LINKED ISSUE #338: [Bug]: ZIP downloads buffer the entire archive in memory [Introduced in 2.1.0]
### Affected Component

### Describe the bug

What happened:
Directory and multi-path ZIP downloads build the **entire** archive in a `bytes.Buffer` and only then send it, so the whole compressed archive is held on the heap before any response byte is written. Combined with the 2 GB upload/pull limits, an authenticated user with default `User`-role resource/flow-file permissions can store (or pull) a large, incompressible directory and request it as a ZIP, forcing the API process to allocate memory proportional to the archive size. A handful of concurrent requests can exhaust process/host memory and crash or severely degrade the API.

Single-file downloads are fine since they stream from disk. Only the directory / multi-path ZIP paths are affected.

WHat should happen:
We should probably create the zip writer over the response writer (`zw := zip.NewWriter(c.Writer)`) after setting `Content-Type: application/zip` and `Content-Disposition`, write entries, then `zw.Close()`.

### Steps to Reproduce

1. As a `User`-role account, upload (or pull from a flow container) a directory of incompressible data.
2. Request the directory (or a multi-path selection) from the ZIP download endpoint.
3. The handler calls `ZipResources` / `ZipDirectory` / `ZipRelativePaths` with a `bytes.Buffer`; resident heap grows to roughly the archive size before the response starts.
4. Issue a few concurrent requests → Go runtime fatal out-of-memory.

A standalone harness that calls the real `ZipResources` with a `bytes.Buffer` sink shows heap growth proportional to archive size (≈32 MiB incompressible input → ≈33.5 MiB ZIP → ≈105 MiB max heap) and crashes with a runtime OOM under a constrained `ulimit -v`, with the stack going through `bytes.growSlice` → `bytes.(*Buffer).Write` → `archive/zip` → `io.Copy` → `resources.ZipResources`.

### System Configuration

- [x] Docker compose

### Logs and Artifacts

_No response_

### Screenshots or Recordings

_No response_

### Verification

sridhar-3009:
Replace `bytes.Buffer` accumulation in `DownloadFlowFile` with direct streaming to `c.Writer` for both the single-directory and multi-path ZIP cases, eliminating the in-memory buffer that could OOM the server for large archives.

Closes #338

### DIFF
--- a/backend/pkg/server/services/flow_files.go
+++ b/backend/pkg/server/services/flow_files.go
@@ -1,7 +1,6 @@
 package services
 
 import (
-	"bytes"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


asdek:
hey @sridhar-3009 

thank you for the PR, but it's already fixed in #339

PR #370, opened 2026-07-09
fix: handle nested backticks in markdown code block rendering

### LINKED ISSUE #361: [Bug]: Triple backticks inside code blocks break markdown rendering
### Affected Component

### Describe the bug

**Description:**

When markdown content contains a code block (fenced with triple backticks) and that code block itself contains triple backticks as part of its content, the inner triple backticks are interpreted as closing the code block instead of being displayed as literal characters. This breaks the entire rendering of the page — everything after the premature code block closure is misinterpreted as markdown, corrupting all subsequent formatting.

This is especially problematic for PentAGI because agents frequently produce output that explains markdown syntax or shows command examples containing triple backticks (e.g. instructions on how to write code blocks, or shell commands wrapped in triple backticks inside documentation).

**Expected behavior:**

Triple backticks that appear inside a fenced code block should be treated as literal text and displayed as-is, not interpreted as code block delimiters. The markdown renderer should properly handle nested or escaped triple backticks.

**Actual behavior:**

Triple backticks inside a code block close the code block prematurely. All content after the inner triple backticks is rendered as regular markdown instead of staying inside the code block, breaking the entire page layout.

### Steps to Reproduce

1. Open PentAGI UI and navigate to any flow where an agent has produced markdown output containing a code block with triple backticks inside it.
2. Observe that the rendered markdown is broken — the inner triple backticks close the code block, and everything after is misformatted.
3. Paste the following markdown into any markdown field in the UI:

```
To run a network scan, the command is ```nmap -sV 10.0.1.50``` then check the results.

The markdown syntax for a code block is: open with ```bash, write your code, then close with ```.

For example, to display JSON:

```json
{"flowId": 85, "action": "finish"}
```

And here is the server response:

```json
{"errors": [{"message": "flow not found"}]}
```

The command ```psql -U pentagi -d pentagi -c "SELECT * FROM flows"``` checks the database state. If the result shows ```waiting``` for a flow that should be ```finished```, it means the provider was renamed.

To fix it, run ```UPDATE flows SET model_provider_name = 'OllamaC GLM' WHERE model_provider_name = 'Ollama cloud GLM';``` then restart with ```systemctl restart pentagi```.
```

4. Observe that the rendering is completely broken — text after the first inner ``` is no longer inside the code block.


### System Configuration

- PentAGI Version: v2.1.0-879e87c (image `vxcontrol/pentagi:latest`)
- Deployment Type: Docker Compose
- Browser: (any modern browser — the issue is renderer-side)


### Logs and Artifacts

No server-side logs — this is a frontend rendering issue.

The bug is reproducible with any markdown content containing backticks. See the example in "Steps to Reproduce" above.


### Screenshots or Recordings

<img width="871" height="603" alt="Image" src="https://github.com/user-attachments/assets/a79fdfa3-97c1-41f3-b927-85fd6ab19cba" />

### Verification

- [x] I have included relevant logs and error messages

sridhar-3009:
Add a `preprocessMarkdownFences()` step in the `Markdown` component that auto-extends the outer fence when a code block body contains a backtick run long enough to prematurely close it, preventing the renderer from splitting a single block into two.

Closes #361

### DIFF
--- a/frontend/src/components/shared/markdown.tsx
+++ b/frontend/src/components/shared/markdown.tsx
@@ -52,6 +52,83 @@ interface MarkdownProps {
     searchValue?: string;
 }
 
+/**
+ * Scan `content` line-by-line for fenced code blocks opened with backtick
+ * fences (```). When the body of a block contains a line whose leading
+ * backtick run is long enough to close the current fence, the outer fence is
+ * extended by one backtick beyond the longest such inner run so the parser
+ * treats the body as literal content instead of closing the block early.
+ *
+ * Only backtick fences are normalised; tilde fences (~) are left untouched
+ * because remark-gfm already handles them correctly.
+ */
+function preprocessMarkdownFences(content: string): string {
+    const lines = content.split('\n');
+    const result: string[] = [];
+    let i = 0;
+
+    while (i < lines.length) {
+        const line = lines[i];
+        // Match an opening backtick fence: optional indent (≤3 spaces), then 3+ backticks, then optional info string.
+        const openMatch = /^( {0,3})(```+)([ \t].*)?$/.exec(line);
+
+        if (!openMatch) {
+            result.push(line);
+            i++;
+            continue;
+        }
+
+        const indent = openMatch[1] ?? '';
+        const fence = openMatch[2];
+        const info = openMatch[3] ?? '';
+        const fenceLen = fence.length;
+
+        // Scan ahead: find the closing fence and track the longest backtick
+        // run on any single line inside the block body.
+        let closeIdx = -1;
+        let maxInnerLen = 0;
+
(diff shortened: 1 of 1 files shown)


hiSandog:
A useful edge-case test would include nested Markdown where the inner backtick run appears inside an indented list or blockquote, since fence recognition can depend on the container prefix. It may also be worth testing a body containing several backtick runs of different lengths and confirming the preprocessor chooses exactly one longer outer fence. Finally, code spans outside fenced blocks should remain byte-for-byte unchanged.

sirozha:
Thanks for the patch. We are not taking this: on the input from the linked issue the rendered structure does not change. Closing this PR.

PR #386, opened 2026-08-03
fix(docker): rebuild flow primary container when it is gone from the daemon

F2had:
## Summary

A flow whose primary container is removed outside PentAGI never recovers. `IsContainerRunning` now reports a missing container as not running, and `Prepare` checks with the daemon before reusing the stored container.

## Problem

A flow stores the Docker id of its primary container in `containers`. Nothing reconciles that row against the daemon, so if the container is removed outside PentAGI (`docker rm`, `docker system prune`, a host rebuild), the row keeps `status = 'running'`.

`flowToolsExecutor.Prepare` trusts that status and reuses the stored id without asking Docker, so every terminal call fails with:

```
terminal tool 'terminal' handled with error: runtime verification failed: container inspection failed: Error response from daemon: No such container: db9eae1e00b7a359059d8fcfa5d2fb1605d32abf5160697fbe6e033d77600ef0
```

Restarting PentAGI does not help, because `Prepare` re-reads the same row. The branch that rebuilds only runs for a status other than `running`, and nothing sets one once the container has gone.

The flow is not marked failed either. `terminal.go:115` returns the failure as tool output with a `nil` error, so the agent treats it as an ordinary result and keeps going, consuming tokens on a flow where no terminal or file operation can succeed.

## Solution

- `IsContainerRunning` returns `(false, nil)` for a container the daemon does not know, instead of wrapping the not-found error. `StopContainer` and `RemoveContainer` already special-case `client.IsErrNotFound` in the same file.
- `Prepare` confirms with the daemon before reusing the stored container. If it is gone, the existing remove-and-rebuild path takes over.

An unreachable daemon is still returned as an error rather than read as "container missing", so a transient failure cannot discard a healthy container.

## User Impact

Flows already broken by this recover on their next load once deployed. No migration or manual database edit.

One behaviour change worth knowing about: an operator who removes a primary container by hand to contain a misbehaving agent will now see it rebuilt on the next flow load. Stopping the flow is the supported way to halt an agent, so this is not gated behind an env var.

## Test Plan

`backend/pkg/docker/client_test.go`: start a container, remove it with `ContainerRemove(..., Force: true)`, then call `IsContainerRunning` with the same id. Before the change it returns the wrapped not-found error; after, `false` with no error, which lets `Prepare` rebuild. A second case covers an id the daemon has never seen.

Both skip when no daemon is reachable, and the first skips rather than pulling when `alpine:3.20` is absent, so neither adds a network dependency.

```
ok  	pentagi/pkg/docker	1.481s
ok  	pentagi/pkg/tools	14.431s
ok  	pentagi/pkg/controller	0.020s
```

`gofmt -l` and `go vet` are clean on the changed files. `go mod tidy` leaves `go.mod` and `go.sum` unchanged.

## Notes

A few lines below this change, `IsContainerRunning` reads `inspection.State.Running` before it checks `inspection.State != nil`. Left alone to keep the diff small.

### DIFF
--- a/backend/pkg/docker/client.go
+++ b/backend/pkg/docker/client.go
@@ -600,7 +600,11 @@ func (dc *dockerClient) Cleanup(ctx context.Context) error {
 func (dc *dockerClient) IsContainerRunning(ctx context.Context, containerID string) (bool, error) {
 	inspection, err := dc.client.ContainerInspect(ctx, containerID)
 	if err != nil {
-		return false, fmt.Errorf("container inspection failed: %w", err)
+		if !client.IsErrNotFound(err) {
+			return false, fmt.Errorf("container inspection failed: %w", err)
+		}
+		// a removed container is missing, not an inspection failure
+		return false, nil
 	}
 
 	// Check both Running state and health status if available
--- a/backend/pkg/docker/client_test.go
+++ b/backend/pkg/docker/client_test.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/client"
 )
 
 func TestStatContainerEntries_AllSucceed(t *testing.T) {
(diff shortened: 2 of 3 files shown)


asdek:
hey @F2had 

thank you for the PR!

PR #390, opened 2026-08-10
feat(i18n): fully localize frontend into Turkish

captainmgc:
## Summary
- Wires i18next/react-i18next across all pages and shared UI components (auth, dashboard, flows, templates/knowledges/resources, settings, data-table, sidebar, dialogs, file manager, markdown editor, etc.)
- Defaults the app language to Turkish
- Keeps `src/locales/en.json` / `tr.json` in sync with full Turkish translations (~800 keys)

## Test plan
- [x] `tsc -b --noEmit` — no new type errors introduced
- [x] `vitest run` — all failures are pre-existing/unrelated (marked package version conflict, missing rxjs transitive dep in sandbox), confirmed identical on `main` before this change
- [x] Verified `en.json`/`tr.json` have identical key sets and are valid JSON


sirozha:
Thanks for the localization work. We are not taking this: it makes Turkish the default language, ships no lockfile, and targets an older state of the frontend. Closing this PR.

sirozha:
Thanks for the localization work. We are not taking this: it makes Turkish the default language, ships no lockfile, and targets an older state of the frontend. Closing this PR.

PR #421, opened 2026-09-15
feat(config): allow skipping tool-loop summaries

SherryOvo:
## Summary

- add `SUMMARIZER_SKIP_TOOL_LOOP_AGENT_TYPES` as an opt-in, comma-separated configuration
- skip incremental `SummarizeChain` calls after tool rounds for matching agent types
- keep restore-time summarization intact as a safety boundary for interrupted chains
- preserve current behavior by default and document large-context-model usage

## Motivation

For tool-heavy specialist agents, an aggressive summarizer configuration can add another LLM request after nearly every tool round. On large-context models this can increase wall-clock latency substantially even when retaining the unsummarized specialist chain is acceptable.

The new option allows operators to choose that tradeoff per agent type instead of disabling summarization globally. Matching is case-insensitive and ignores surrounding whitespace.

Example:

```env
SUMMARIZER_SKIP_TOOL_LOOP_AGENT_TYPES=coder,searcher,enricher,installer,pentester
```

Leaving the variable empty retains the existing behavior.

## Validation

- `go test ./pkg/config ./pkg/providers ./pkg/csum -count=1`
- `go build ./cmd/pentagi`
- `git diff --check`


sirozha:
Thanks for the patch. We are not taking this: when the model window is unknown, that summarization is the only bound on chain growth. Closing this PR.

